Issue #175, opened 2019-01-23
age key is jumping off page when looked at on small computer screen


polly64:

tomalrussell:
Max height of legend needs limiting on small-height screens (possibly using `vh` units instead of `rem`? or add a media-query?).

Here's a screenshot of the problem:
![image](https://user-images.githubusercontent.com/2762769/53800316-5ba68600-3f34-11e9-93a4-49d16307e437.png)

Fix should be CSS-only - https://github.com/tomalrussell/colouring-london/blob/master/app/src/frontend/legend.css#L36

martin-dj:
Fix submitted to feature branch which seems to have resolved the issue on small screens.

tomalrussell:
Closed by #225

PR #225, opened 2019-03-14
Fix for issue #175-age key is jumping off page when looked at on smal…

martin-dj:
…l computer screen. Set max-height to 50 vh.


tomalrussell:
Closes #175

PR #386, opened 2019-08-14
Refactor API and modify routes

### LINKED ISSUE #383: Refactor API/frontend routes and handle page/resource not found
As discussed in #323: move out the JSON API under its own endpoint `/api`, improve 404 responses (prevent raw `Cannot GET /something` responses from frontend)

mz8i:
Resolves #383


mz8i:
There are some changes to the frontend code, in order to adapt to the new API routes, but the URL values are the only thing changed.

app/src/api/api.ts
```diff
@@ -0,0 +1,141 @@
+import express from 'express';
+import bodyParser from 'body-parser';
+
+import { authUser, createUser, getUserById, getNewUserAPIKey } from './services/user';
+import { queryLocation } from './services/search';
+
+import buildingsRouter from './routes/buildingsRouter';
+
+
+const server = express();
```
tomalrussell:
```suggestion
const server = express.Router()
```

mz8i: This was kind of intentional as I feel long term we should gravitate towards the web app backend and the data API being two separate applications (note that the use of `body-parser` was moved to only be applied in `api.ts` as we don't use JSON requests in the frontend serving part of the app), but actually since we don't start them separately, a Router is probably fine for now.

tomalrussell: Makes sense as a design idea - at some point back in time I had the front end, tileserver and API server split out as separate, but YAGNI and deployment simplicity won out, especially before we have data on performance and usage stats - if any request can be handled by one kind of process then we can simply scale that process and load should be spread simply across them.

PR #390, opened 2019-08-15
Feature 63: delete user account (backend code)

mz8i:
Relies on #386 being merged first.
Please note this will be followed by a second PR for the frontend part of the user delete functionality.

PR #392, opened 2019-08-16
Feature 63: delete user account (frontend code)

### LINKED ISSUE #63: Delete account
- [ ] Delete account form, linked from my-account
- [ ] Double-check that they want to delete!
- [ ] User delete API endpoint (must be authenticated)
- [ ] Remove user details (email/password/registered) from database
- [ ] ~~Create unique username? `deleted_user_123`~~

Make clear that this will not remove references from the edit log - be clear that it's not possible to retract contributions, that old archives with username will continue to be publicly available.

mz8i:
Resolves #63.
Contains frontend code for creating a modal confirmation popup, and changes to my-account view that add a Delete account button and ask the user to confirm the choice.

The styling of the popup is based on the Bootstrap modal classes, but the mechanism for showing and hiding the modal is added on separately - this IMHO is not as good as using something like a React-Bootstrap component, which relies also manages stuff like aria attributes. We might want to think about using React-Bootstrap?

The changes from the backend PR (#390) seem to also be included in this PR - not entirely sure about this but might be because we modified the commits for the backend PR slightly.


app/src/frontend/ConfirmationModal.css
```diff
@@ -0,0 +1,7 @@
+.modal.modal-show {
```
tomalrussell: Niggle comment on filename - please stick with the hyphen-case-convention - so `confirmation-modal.css`

mz8i: ok! will correct

Issue #61, opened 2018-09-14
Reset password flow


tomalrussell:
- [ ] View reset password request form (email address)
- [ ] Generate token
- [ ] Send email (look at mailgun)
- [ ] Validate token
- [ ] View reset password form (new password/confirm)
- [ ] Update password + login

mz8i:
@tomalrussell If the user registered without an e-mail, then it's not possible to reset a password, right?

mz8i:
Re mailgun, do we actually want to use that platform?

tomalrussell:
Can you flesh out pros/cons/options?

- Simply sending from the server
- we do have a mailgun account already; the discourse forum uses it for transactional emails
- data protection considerations?
________________________________

mz8i:
If we already have an account then I think it makes sense to use it! I'll read up on how to set up Node as a client

mz8i:
I think we should use Mailgun but over SMTP rather than using their proprietary API, and use some popular library like Nodemailer on the Node side - this way we only need to change the endpoint information if we ever want to switch for example to a self-hosted SMTP server.

dominijk:
Lfgss.com use a service (mircosm?) which works well based on my experience of using it for a few years. I don’t have a password just an email,  when I clear my cache or login elsewhere i just enter my email and get a new PIN number emailed which the client holds onto

tomalrussell:
@mz81 - :+1: making it easy to switch provider, mailgun+smtp+nodemailer sounds like a good approach

@dominijk :+1: email-only login - should be covered by #62

PR #399, opened 2019-08-21
Feature 61: forgotten password

mz8i:
Work towards #61.

The current commits contain the backend part of the password reset functionality. Password reset can already be carried out through sending requests to the API. Work on frontend in progress.


mz8i:
Will need to improve the mechanism for assembling the link in password reset email, due to https://www.skeletonscribe.net/2013/05/practical-http-host-header-attacks.html

app/src/api/controllers/userController.ts
```diff
@@ -0,0 +1,105 @@
...
+    throw new Error('adsd');
+    if(req.body == undefined || (req.body.email == undefined && req.body.token == undefined)) {
+        return res.send({ error: 'Expected an email address or password reset token in the request body' });
+    }
+
+    if(req.body.email != undefined) {
+        // first stage: send reset token to email address
+
+        // this relies on the API being on the same hostname as the frontend
+        const { origin } = new URL(req.protocol + '://' + req.headers.host);
```
tomalrussell: In absence of brighter ideas, this could come from an environment variable (to avoid trusting request headers).

app/src/api/controllers/userController.ts
```diff
@@ -0,0 +1,105 @@
...
+        () => userService.logout(req.session)
+    ).then(() => {
+        res.send({ success: true });
+    }).catch(err => {
+        res.send({ error: err });
+    });
+}
+
+async function resetPassword(req: express.Request, res: express.Response) {
+    throw new Error('adsd');
```
tomalrussell: Always fails - for testing?

app/src/api/controllers/userController.ts
mz8i: My thoughts too. I wanted it to be automatic but I guess updating one variable between dev and prod is not too much of a hassle. I would suggest having `WEBAPP_ORIGIN` then, set to for example `http://localhost:3000` or `https://beta.colouring.london` etc?

app/src/api/controllers/userController.ts
mz8i: ah, sorry about that... Yes indeed, not sure how I managed to leave it there. Pushing a fix now.

PR #403, opened 2019-08-23
[WIP] Feature/view edit refactor

tomalrussell:
This currently breaks the multi-edit view, but is otherwise close to getting in shape.

Big idea: declarative react view for each "category" 

This moves away from the json fields-config to plain-old components.

- `App` routes to 
- `BuildingView` which switches on category, rendering
- e.g. `LocationContainer` which contains all the data-entry inputs, with names/labels/tooltips - most of the fields-config.json reworked directly into component attributes.

To reuse copy/edit logic, and header/wrapper HTML elements, each `***Container` (in `frontend/building/data-containers`) is wrapped by `withCopyEdit` from `frontend/building/data-container.tsx`. This is the [higher-order components](https://reactjs.org/docs/higher-order-components.html) pattern to compose the simple `***Container` with the stateful wrapping that handles switching modes and persisting data.

Individual field components are all in `frontend/building/data-components`. The idea is that these can grow too, to allow for friendlier user experience but exposing a simple interface to components higher up the tree. For example, `MultiDataEntry` is a small stateful component that wraps around a list of text inputs. `YearDataEntry` is a sketch towards having simpler date selection, with a few TODOs.


app/src/frontend/building/categories.tsx
```diff
@@ -0,0 +1,147 @@
...
+const Categories = (props) => (
+    <Sidebar>
+        <Category
+            title="Location"
+            slug="location"
+            help="https://pages.colouring.london/location"
+            inactive={false}
+            mode={props.mode}
+            building_id={props.building_id}
+        />
```
mz8i: Nit-pick: could just do `{...props}` in each `Category` to reduce repetition?

tomalrussell: I'm less keen on passing `{...props}` through since realising it makes it easy to rely on things implicitly which may be set up two or more layers away - "Explicit is better than implicit" in a Zen-of-Python kind of way.

mz8i: True. What could reduce the implicitness would be having TS interfaces defined for prop types of both components - then if a field was added to the prop types of Categories but not Category, it would result in a transpilation error, because the destructuring process would try to assign to properties that do not exist on Category.

tomalrussell:
> In what ways is the creation of special components for categories and data fields better than having all the definitions in the JSON file? 

It drops a layer of abstraction which worked neatly when we had a small set of input types (number/text/select) and it looked like we might want to quickly add new fields in each category without much further customisation. 

Now we're looking at more custom handling of inputs and introducting some hierarchy or nested components - dropping the custom JSON mini-specification in favour of plain-old-react-components should make it more direct to add, extend and customise the sidebar UI.

> How much does it increase the effort necessary to change some information across the site? It would seem to me that there will be quite a few places one has to edit? 

If we want to add a field to a category in the sidebar UI, it's a change in one place, as before - the relevant data-container (say `app/src/frontend/building/data-containers/community.tsx`). We still need to add the database field, and whitelist and validate it in the API and building service.

If we want to use a new custom control, then we can add a new component to  `app/src/frontend/building/data-components` and use it. Because we now share data-components between view and edit modes, there's a bit less duplication.

> I understand that one upside is being able to easier define some more advanced interactions during data input?

That's the intention, yes - if we want to introduce layers of hierarchy or more complicated data-entry components that guide users through the process a bit more, we can do that fairly directly, at the level of the `data-components`. Depending on just how complex these become, we might need to slice up or reorganise further - hopefully this is a reasonably well-structured point to allow for further change later.

PR #408, opened 2019-08-30
Features/migrations team

### LINKED ISSUE #362: db: Add Team category awards fields (type & year) to database
**Fields for launch**
- Number of awards for building (can activate, set min max of ?20?) - (map colour with be as location, based on number of awards)
- Award names - award naming body 
- Year of award (with min max constraint from year of earliest award to max of current year) 

**Steps for adding new fields to database;**

- [x] Identify bulk data sources (optional)
- [x] Required - Define field values as static category page using [this](https://awards-list.co.uk/uk-awards/construction-awards-and-architecture-awards/) as initial list
- [ ] Required - Identify need for source and verification for fields, log to spreadsheet
- [x] Required - Define input method for field (Drop down with multi select and auto complete) 
- [x] Required - Add category down file
- [x] Required - Add category up file
- [x] Required - If not all fields added create new issue and link, copy steps above #406 
- [x] On completion set up Github issue for activating fields on front end #407

dominijk:
This will close #362, I've opted for `ENUM` types for award names as discussed in issue


migrations/013.team.up.sql
```diff
@@ -0,0 +1,44 @@
+-- Remove team fields, update in paralell with adding new fields
+--Numeric small int, if > 1 triggers multifield boxes on fron end
+ALTER TABLE buildings
+    ADD COLUMN IF NOT EXISTS number_of_awards smallint;
+
+-- Building award, can be any of:
+-- This is an edit of all potential awards to the most commonly recognised with bias to community, hertiage and design focussed ones
+CREATE TYPE team_awards
+AS ENUM ('Architects Jorunal - Heritage Project',
+    'Architects Jorunal - Housing Project',
```
tomalrussell: Typo for Journal (Jorunal) here and below?

migrations/013.team.up.sql
```diff
@@ -0,0 +1,44 @@
...
+    'Neave Brown Award',
+    'New London Architecture awards - Conservation and retrofit',
+    'New London Architecture awards - Mayors Prize',
+    'New London Architecture awards - Culture & Community',
+    'New London Architecture awards - Education',
+    'New London Architecture awards - Homes',
+    'New London Architecture awards - Housing',
+    'New London Architecture awards - Offices',
+    'RIBA Gold mdeal',
+    'RIBA Sitrling prize',
```
tomalrussell: Typo for Stirling?

migrations/013.team.up.sql
```diff
@@ -0,0 +1,44 @@
...
+    'ICE London Civil Engineering Awards',
+    'Neave Brown Award',
+    'New London Architecture awards - Conservation and retrofit',
+    'New London Architecture awards - Mayors Prize',
+    'New London Architecture awards - Culture & Community',
+    'New London Architecture awards - Education',
+    'New London Architecture awards - Homes',
+    'New London Architecture awards - Housing',
+    'New London Architecture awards - Offices',
+    'RIBA Gold mdeal',
```
tomalrussell: Typo for medal?

dominijk:
Potentially its multiple awards per building I'm not sure how often that happens but the norm would be single award per building, option 2 makes sense to me. We've not set one of those up yet have we? It's new to me so I'll go find some docs.

tomalrussell:
I think so far we use jsonb only in the logs table to store patches: [001.core.up.sql#L126](https://github.com/tomalrussell/colouring-london/blob/master/migrations/001.core.up.sql#L126)

References:
- the data type: https://www.postgresql.org/docs/current/datatype-json.html
- functions and operators: https://www.postgresql.org/docs/current/functions-json.html
- blog post intro: https://blog.codeship.com/unleash-the-power-of-storing-json-in-postgres/

dominijk:
Have you got any experience on validating input for `jsonb ` in postgresql? There seem to be a couple of options to validate on the db that it's a valid object 
- https://stackoverflow.com/questions/40096394/is-there-a-way-to-enforce-a-schema-for-a-json-field-within-a-postgresql-record-i

What t I'm trying to figure out if there's something that can look inside the object to check the content is acceptable that isn't a function as that seems an expensive operation to run. Or does that happen in the client? I'm thinking in n particular the free text problem. If I understand correctly in the client I can apply some restrictions that state it has to be an integer, is their an equivalent to `ENUM` if so where would the list of possible awards reside?

Issue #342, opened 2019-08-02
API: Bulk download of file 


dominijk:
Enable users to download simple flat file of data in CL
- [ ] extract open dataset to archive storage (current state + edit history)
- [ ] simple list-of-links page to download "Latest" and "Archive" bulk extracts

mz8i:
@tomalrussell what format do we want to export the edit history in?
Specifically:
- do we include forward (and reverse?) patch or some other description of changes?
- for edit author, do we want user's UUID or name (potentially modified over time by user deletion)
- I assume bulk edits are split into individual building edits? Given that there's a `building_id` in `logs`

mz8i:
On the topic of IDs / UUIDs, is there any case at all where we want to include the internal database IDs in the export files? There's many - `building_id`, `geometry_id`, `revision_id`, `building_property_id`, `log_id` etc

tomalrussell:
Include where they're necessary for cross-reference between exported tables:
- `building_id` is essential
- `geometry_id` skip
- `revision_id` include as cross-reference to last edit in history
- `building_property_id` skip
- `log_id` include for reference by `revision_id`

mz8i:
I'd think about renaming either `revision_id` or `log_id` - at least in the exported files. I overlooked the foreign key so didn't realise they were referring to each other, but to avoid confusion for users of the extract I'd name them the same

mz8i:
Closed by #411

PR #411, opened 2019-08-30
Feature 342: bulk extracts

mz8i:
Work towards #342 - contains python export script, cron schedule, API back-end, and front-end.

Downloads page is accessible under `/data-extracts.html` but there's no link to that page currently - needs discussion where to link from.

still missing: nginx set up to serve archive files (the download links won't work right now)

To test, run migrations and define local path in EXTRACTS_DIRECTORY env var when running python script.


maintenance/extract_data/extract_data.py
```diff
@@ -54,7 +54,7 @@ def db_to_csv(connection, query):
 
 def get_extract_zip_file_path(current_time):
     base_dir = Path(os.environ['EXTRACTS_DIRECTORY'])
-    file_name = f"data-extract-{current_time:%Y-%m-%d}.zip"
+    file_name = f"data-extract-{current_time:%Y-%m-%d-%H_%M_%S}.zip"
```
mz8i: The time is included in the filename to avoid overwriting a file, since that would create two database records pointing to the same file.
While the above should fix the problem because I don't envision two extracts being created in the same second, I'm thinking it might actually be worth throwing an error if we're trying to overwrite a file that already exists, because having two extract records with the same file is a bad situation

app/src/api/services/dataExtract.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export async function listDataExtracts(): Promise<DataExtract[]> {
+    try {
+        const extractRecords = await db.many<DataExtractRow>(
+            `SELECT
+                extract_id, extracted_on, extract_path
+            FROM bulk_extracts`
+        );
+
```
mz8i: To consider: apparently `pg-node` converts a `timestamp without timezone` type to a `Date` in the server's local time zone, while we store timestamps in UTC in the database. We probably need to convert the result adequately, a workaround is mentioned here: https://stackoverflow.com/a/38340396

maintenance/tasks.cron.txt
```diff
@@ -0,0 +1 @@
+0 5 * * * /var/www/colouringlondon/maintenance/extract_data/extract_data.py
```
mz8i: 1. AFAIK the path needs to job script needs to be absolute so I put the standard path to the project on our servers
2. This file could be imported with `cat tasks.cron.txt | crontab -e -` but that would override any other jobs that are not in this file. The solution could be:
a) version control all cron tasks we want to run on the server in the `tasks` file (more comprehensive but might lead to exposing some details that are specific only to our server set-up, like the back-up schedule)
b) do not store cron schedules in version control (but then the schedules, which are sometimes tied to the proper functioning of the website, could get lost / accidentally overwritten etc)

maintenance/extract_data/extract_data.py
tomalrussell: Including timestamp looks good; yes let's throw an error.

app/src/api/services/dataExtract.ts
tomalrussell: From the stackoverflow it looks like the best workaround would be to store `timestamp with timezone` explicitly as UTC.

maintenance/tasks.cron.txt
tomalrussell: 1. Looks fine
2. Do you think we should move to version-controlled config in a private repo? I.e. option a) but avoiding exposing unnecessary detail.

app/src/frontend/data-extracts.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+        this.state = {
+            latestExtract: undefined,
+            previousExtracts: undefined
+        };
+    }
+
+    async componentDidMount() {
+        const res = await fetch('/api/extracts');
+        const data = JSON.parse(await res.text(), dateReviver);
```
tomalrussell: Using `JSON.parse` with `dateReviver` to create `Date` objects, which `await res.json()` wouldn't do?

app/src/frontend/data-extracts.tsx
```diff
@@ -0,0 +1,70 @@
...
+                            <div>
+                                <h1 className="h3">Latest extract</h1>
+                                <ExtractDownloadLink {...this.state.latestExtract} />
+                            </div>
+                    }
+                    {
+                        this.state.previousExtracts && this.state.previousExtracts.length > 0 ?
+                            (<div>
+                                <h1 className="h3">Older extracts</h1>
+                                {this.state.previousExtracts.map(e => <ExtractDownloadLink {...e} />)}
```
tomalrussell: Could be good as a list (`ul`)?

app/src/api/routes/asyncController.ts
```diff
@@ -0,0 +1,8 @@
+import { Request, Response, NextFunction } from 'express';
+
+export default function asyncController(fn: (req: Request, res: Response, next: NextFunction) => Promise<any>) {
```
tomalrussell: This needs a block comment to explain the intention - wrap async controller functions to be used as express routes.

app/src/api/services/dataExtract.ts
```diff
@@ -0,0 +1,59 @@
...
+    extract_path: string;
+}
+
+interface DataExtract {
+    extract_id: number;
+    extracted_on: Date;
+    download_path: string;
+}
+
+export async function listDataExtracts(): Promise<DataExtract[]> {
```
tomalrussell: Here and elsewhere - let's follow the convention of exporting at the bottom of a file in one place.

app/src/frontend/data-extracts.tsx
tomalrussell: (I could have used a one-line comment just to point this out as it's less familiar)

mz8i: I don't quite understand what you mean here, could you rephrase please?

mz8i: ah I think I understand now. Yes, this is to actually parse the JSON date strings into Date objects. Sure, I will aim to put comments in such places where it's slightly less obvious what's happening. Do you want them in the code or rather in a comment to the PR?

app/src/api/services/dataExtract.ts
mz8i: to be honest I am now a bit confused about the handling of time zones in Postgres. From what I've just read in the documentation, `timestamptz` stores UTC values in the database, but upon reading them it assumes that they are in the server's local time zone. What do you then mean, in this context, by storing the dates explicitly as UTC?

maintenance/tasks.cron.txt
mz8i: I would say versioning the configuration but separately from the source would be the way to go. Can you please create the private repo then?

mz8i:
@tomalrussell re the explicit model interfaces, I agree it'd be nice to start annotating more types like that, although we should be more systematic than I was here. I can identify at least three types of data models - ORM models for mapping to database rows (see `DataExtractRow`), internal model for working with on the backend (see `DataExtract`) and the model for the frontend (see `ExtractViewModel`). Here, the latter two are identical. Could potentially have them as one interface.

I think I am happy with the performance for now, but it might be something to monitor. Especially the edit history will be presumably growing quite quickly.

app/src/api/services/dataExtract.ts
mz8i: I guess you mean setting the server time zone explicitly to UTC, upon which `timestamptz` should always be treated as UTC when querying the database.

mz8i: So it seems indeed that the best practice is to keep most columns as `timestamp with time zone` (`timestamptz`). Right now, all our columns in the database are `timestamp without time zone`, and the current time zone set on the production and staging servers is `Europe/London`. I suggest that as part of this PR, we adjust the type of all columns (following [this SO answer](https://dba.stackexchange.com/a/134603) ), change the server time zone to `UTC`, and verify that all the code dealing with timestamps is correct given the new types.

app/src/frontend/data-extracts.tsx
tomalrussell: Great :) yes, I meant that I was briefly puzzled by why the `dateReviver` function was necessary. A short explanatory comment in the code would help.

app/src/api/services/dataExtract.ts
tomalrussell: Suggestion sounds good :+1:

mz8i:
nginx config was added to private config repo. Files will be served from URL `/downloads/*.zip` and will serve files from server.
Not setting scheduled export yet - for launch will only run the export manually.
Link to downloads page added to header.

PR #417, opened 2019-09-09
UI/UX Improvements

mz8i:
A wide array of changes regarding UX/UI, mostly mobile but also general.
Depends on commits from https://github.com/tomalrussell/colouring-london/tree/feature/view_edit_refactor

UPDATE 2019/09/17:
Main changes included in this PR:
- refactor app routing, move out all map-related screens to their own `map-app.ts` component with nested routing regarding the sidebar/welcome
- persist last chosen category on map
- allow no building selected, allow deselecting current building by clicking on the selected building or on an area with no building
- use flex to lay out the page, improve styles of several elements
- map takes up only visible space, does not display under sidebar etc. All map controls are positioned relative to map viewport.
- improve legend styling/sizing/collapsing

Known remaining issues:
- info icon in category tiles does not display on mobile
- updating map after saved edit doesn't work because of the change in building de-selection mechanism


app/src/frontend/pages/welcome.css
```diff
@@ -6,8 +6,10 @@
     z-index: 10000;
     top: 0;
     width: 100%;
+    max-height: 100%;
     border-radius: 0;
     padding: 1.5em 2.5em 2.5em;
+    overflow-y: scroll;
```
tomalrussell: I sometimes confuse `overflow: auto` with `scroll` - does `scroll` force a (greyed-out) scrollbar in some browsers, even when there's no overflow? I think `auto` is the one that will enable scrolling if necessary and have no scrollbar otherwise.

mz8i: I think auto will be more appropriate, I was even thinking about it today after I saw that in Chrome

app/src/frontend/map-app.tsx
```diff
@@ -0,0 +1,250 @@
...
+            building: props.building,
+            building_like: props.building_like
+        };
+
+        this.selectBuilding = this.selectBuilding.bind(this);
+        this.colourBuilding = this.colourBuilding.bind(this);
+        this.increaseRevision = this.increaseRevision.bind(this);
+    }
+
+    componentWillReceiveProps(props: Readonly<MapAppProps>) {
```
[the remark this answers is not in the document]
mz8i: Normally you wouldn't need to store state derived from props, but here it's used to store the last category - when going back to `/view/categories` from `/view/location`, the category prop would change to `null` and that's what this handler takes care of - it won't display null (or default = age) category, but rather stick to the previous one stored in the state.

Fair point that the handler is deprecated. I looked briefly at the proposed solutions but my knowledge of React is rather basic so far, and I'll have to read a bit more to know which to choose.

mz8i:
> Lot's of nice work, quite a quick review here - see also couple of small comments below.
> 
> Can you take a look at the `category` state variable and see if we can avoid `componentWillReceiveProps`?
> 
> I can see the motivation for pushing anything to do with buildings down into `MapApp`. Does this affect the flow of state changes when user interactions change a category or building? What's the sequence of triggers and events? Do you have any ideas about how best to structure this any more clearly than we're doing now?

Re your last point, I am not entirely sure what you mean. The general mechanism doesn't actually change that much, the main difference is that before we had two parallel Switches directly in the App, and now one of them is actually nested inside the MapApp. I am also avoiding having to parse the URL manually anywhere, and relying on React Router to provide all category- and building-related params to the MapApp. And finally, the ColouringMap becomes quite a simple display component, which doesn't need to know about its context too much.
Maybe if you can give an example of what you mean, I could give a better response?

PR #424, opened 2019-09-17
Feature 286: validate username

### LINKED ISSUE #286: Duplicate user names
I registered on desktop at casa with casa_demonstration as a user name, I then entered the same details in laptop and it didn't recognise them. I then registered on the laptop under the same username and it accepted the registration. This means there should be two users names registered as casa_demonstration one with an email address attached one without

mz8i:
Resolves #286 
Relies on #417 being merged first
Please double-check validation rules in `app/src/api/validate.ts`


mz8i:
That's good advice, I thought we'd need some JS client-side validation for proper pattern validation but according to the linked resource, if you set a `title` attribute on the text input, it will also show up in the failed validation popup, so the user knows what the pattern is. It works quite nicely.
I also increased max password length to 128 characters as per OWASP Authentication cheat sheet: https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Authentication_Cheat_Sheet.md#implement-proper-password-strength-controls

PR #443, opened 2019-10-02
Fix/like

tomalrussell:
Enables ticking 'like' to like a building


mz8i:
Looks simple enough, although I am not fully familiar with the workings of the withCopyEdit HOE.
As a side note, I have run into a similar bug where SelectDataEntry was using props.handleChange where it should be using props.onChange. This prevented dropdowns from being editable. The fix will be released with the sustainability category activation feature.

PR #451, opened 2019-10-02
Fix/category default values

mz8i:


tomalrussell:
While we're here - @all-contributors please add mz8i for code and ideas!

PR #457, opened 2019-10-03
Remove unused 'configuration' field in size

### LINKED ISSUE #409: UI/UX: Size & shape category edit
Move Configuration drop down from this category to Type category

tomalrussell:
Closes #409

Building attachment/configuration is now in type


mz8i:
Can merge as far as I am concerned, but maybe to `develop`? We need to work out a strategy of working with these two branches.

PR #463, opened 2019-10-07
Simplify tile renderer, improve stitching and caching

### LINKED ISSUE #438: Refactor tile server code to be more function-oriented
Following #423, this is to further refactor the new tile server code to move away from a class-oriented, and towards a more function-oriented, renderer composition system.

mz8i:
Resolves #438, and contains the following fixes/improvements:
- tile stitching now uses sharp's `composite()` rather than multiple `overlayWith()` calls, which reduces the number of times a `.png().toBuffer()` sequence is called from 5 to 2. Without precise testing, on a local server this seems to result in a performance improvement at lower zoom levels. An issue was created against sharp (https://github.com/lovell/sharp/issues/1908) related to a problem with compositing and then resizing which seemingly prevents us from being able to complete the operations in just one pipeline.
- the re-written tileserver contained an unintended difference comparing to the old implementation, in that the `base_light` and `base_night` layers were not being cached. This has now been amended and a mechanism added to not clear these layers on building data edit


mz8i:
Apparently sharp requires to create two processing pipelines in the stitching step (first composite, then resize) so this is the best we can get.

Issue #335, opened 2019-08-02
Map: Add GLA boundary to site


dominijk:
As discussed at meeting 1st Aug 2019, this will be a GeoJSON object. @tomalrussell  There are two file options for this  

- [ ] Simplified file (48kb) as stored in [repo](https://github.com/tomalrussell/colouring-london/blob/master/docs/GLA_BoundaryWGS84.geojson)
- [ ] Full [resolution](https://data.cdrc.ac.uk/dataset/london) file (279kb)

mz8i:
Feature completed, need follow-up work on including attribution

PR #464, opened 2019-10-08
Add GLA boundary to map

mz8i:
Work towards #335 

Added both boundary files to static files. Used the detailed one for now. Need discussion on attribution: http://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/

We could have a more sophisticated mechanism for loading the boundary (none, simplified or detailed) but the 272KB download is actually negligible compared to all the map tiles.

There is currently one boundary colour (`#bbb`) for both light and dark themes. I tried to choose it so that it doesn't pop out of the screen too much, but maybe we should differentiate by theme and make the dark theme boundary colour a bit darker. Comments welcome.

image
image


mz8i:
@polly64 could you take a look if the boundary colours from the above screenshots look good to you?

polly64:
yes great

PR #465, opened 2019-10-08
Feature 300: field groups

### LINKED ISSUE #237: UX: Nest categories in 'size'
E.g.

- floor area
  - ground
  - total

- storeys
  - core
  - attic
  - basement

### LINKED ISSUE #300: UX: Hierarchy in planning category panel
Expand/collapse planning data fields, in two main groups:
- everything from 'Is it in a conservation area?' to 'Historic area assessment link' as a group under 'Listing and protections'
- the demolition-related fields under 'Demolition and demolition history'

mz8i:
Resolves #300, resolves #237 (after edit)
Adds a collapsible component for data entry groups, and adds the hierarchy to the planning category.

### DIFF
--- a/app/src/frontend/building/data-components/data-entry-group.tsx
+++ b/app/src/frontend/building/data-components/data-entry-group.tsx
@@ -1,46 +1,34 @@
-import React, { Fragment } from "react";
+import React, { Fragment, useState } from "react";
 
 import './data-entry-group.css';
 import { RightIcon, DownIcon } from "../../components/icons";
 
-interface DataEntryProps {
+interface DataEntryGroupProps {
+    /** Name of the group */
     name: string;
+    /** Whether the group should be collapsed initially */
+    collapsed?: boolean;
 }
 
-interface DataEntryState {
-    collapsed: boolean;
-}
+const DataEntryGroup: React.FunctionComponent<DataEntryGroupProps> = (props) => {
+    const {collapsed: initialCollapsed = true} = props;
 
-class DataEntryGroup extends React.Component<DataEntryProps, DataEntryState> {
-    constructor(props) {
-        super(props);
+    const [collapsed, setCollapsed] = useState(initialCollapsed);
 
-        this.state = {
-            collapsed: true
-        };
-
-        this.toggle = this.toggle.bind(this);
-    }
-
-    toggle() {
-        this.setState((state) => ({
-            collapsed: !state.collapsed
-        }));
-    }
-
-    render() {
-        return (
-            <Fragment>
-                <div className='data-entry-group-header' onClick={this.toggle}>
-                    <CollapseIcon collapsed={this.state.collapsed} />
-                    <span className='data-entry-group-title'>{this.props.name}</span>
-                </div>
-                <div className={`data-entry-group-body ${this.state.collapsed ? 'collapse' : ''}`}>
-                    {this.props.children}
-                </div>
-            </Fragment>
-        );
-    }
+    return (
+        <Fragment>
+            <div className='data-entry-group-header' onClick={() => setCollapsed(!collapsed)}>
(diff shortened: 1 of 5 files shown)


app/src/frontend/building/data-components/data-entry-group.tsx
```diff
@@ -0,0 +1,54 @@
...
+    render() {
+        return (
+            <Fragment>
+                <div className='data-entry-group-header' onClick={this.toggle}>
+                    <CollapseIcon collapsed={this.state.collapsed} />
+                    <span className='data-entry-group-title'>{this.props.name}</span>
+                </div>
+                <div className={`data-entry-group-body ${this.state.collapsed ? 'collapse' : ''}`}>
+                    {this.props.children}
+                </div>
```
mz8i: Wasn't entirely sure if the header and body should be grouped in another div or not?

mz8i:
I changed the DataEntryGroup component to a functional one using the new feature of React hooks - just wanted to experiment after reading the tutorial. Can easily revert to class if preferred, but the React docs encourage to mix the styles and try out hooks on a couple smaller components.

Also, added UI for count of attributes in the group (so more info is available when group is collapsed) and added prop to have group expanded at first.
Used this in Size & Shape category so this PR now also resolves #237

mz8i:
Also, a problem that arises with these groups: when selecting a new building, all the groups go back to their original state (usually collapsed, unless specified otherwise).
This is because the building view component gets recreated. We need to rethink the lifecycle of the components in the sidebar...

tomalrussell:
Can you think through that design problem as part of this bit of work?

What's the immediately simplest solution - to hold the un/collapsed state at a higher level?

Is the app growing to the point that a state-management library like [redux](https://react-redux.js.org/) would reduce complexity?

mz8i:
I'll take a deeper look at this. Storing the un/collapse state at a higher level doesn't feel right, as this is a purely UI thing particular to the category view.

I will take a look at possible solutions. Having not used redux I'll have to do some research into the potential. A recent built-in alternative for state management is React's Context API

mz8i:
After research on Redux, Context API and related concepts, I don't feel we need Redux just yet. The app's state is still mostly just one selected building.
If I understand correctly, the problem with the collapsing groups is because we use `key=<BUILDING_ID>` on the withCopyEdit containers to display new data when another building is selected. But that always remounts the whole subtree so the state of the field group components is reset.
The solution I'd try is to not use the `key` mechanism, and modify withCopyEdit to manage the sync between its props and state more directly. It could perhaps be simpler for the state of withCopyEdit to only contain the currently modified fields - and at render time, merge the props with the modifications. It would also make it easier to check if there are any unsaved changes. I would maybe lift that piece of state (are there any unsaved changes?) up to the MapApp, so that a prompt can be displayed before navigation - this feature is currently missing.

If the general form of the field groups feature is OK, I'd say let's merge it into `develop` and I'll create the above modifications on a separate branch.

tomalrussell:
> The solution I'd try is to not use the `key` mechanism, and modify withCopyEdit to manage the sync between its props and state more directly. It could perhaps be simpler for the state of withCopyEdit to only contain the currently modified fields - and at render time, merge the props with the modifications. It would also make it easier to check if there are any unsaved changes. I would maybe lift that piece of state (are there any unsaved changes?) up to the MapApp, so that a prompt can be displayed before navigation - this feature is currently missing.

Sounds like a useful bundle of work. 

The set of modifications could also be used when saving a building - the API already accepts a subset of building fields (essentially a [JSON-merge-patch](https://tools.ietf.org/html/rfc7396)).

> If the general form of the field groups feature is OK, I'd say let's merge it into `develop` and I'll create the above modifications on a separate branch.

Go ahead.

PR #472, opened 2019-10-15
Feature: store data edits in data container state

mz8i:
This is a WIP for now, as still extending the new mechanism.
This PR bundles changes that were carried out while working on the withCopyEdit component. All are related to the view/edit/copy etc workflow


mz8i:
The current version contains most functionality needed to make this feature work nicely.

Some further improvements we might want to consider at a later date:
- re-organise the links in the header (suggestion: always show info as leftmost, view/edit as rightmost, copy-related links in the middle)
- potentially move save/discard buttons to header (e.g. disable view, edit or save+discard buttons depending on state)
- only allow copy when there are no unsaved changes, disable editing during selection of fields to copy (this might not actually be desirable)
- prompt user when trying to navigate while unsaved changes are present

app/src/frontend/building/building-view.tsx
```diff
@@ -14,26 +14,35 @@ import StreetscapeContainer from './data-containers/streetscape';
...
  *
  * @param props
  */
-const BuildingView = (props) => {
+const BuildingView: React.FunctionComponent<BuildingViewProps> = (props) => {
     switch (props.cat) {
         case 'location':
             return <LocationContainer
                 {...props}
-                key={props.building && props.building.building_id}
```
mz8i: this and other similar lines in this file: the whole component subtree is no longer recreated on building change. This prevent UI-only-related state (like field groups being collapsed or not) from being reset.

app/src/frontend/building/data-components/data-entry.tsx
```diff
@@ -20,7 +37,13 @@ const DataEntry: React.FunctionComponent<any> = (props) => { // TODO: remove any
                 maxLength={props.maxLength}
                 disabled={props.mode === 'view' || props.disabled}
                 placeholder={props.placeholder}
-                onChange={props.onChange}
+                onChange={e => {
+                    const transform = props.valueTransform || (x => x);
+                    const val = e.target.value === '' ?
+                                    null :
+                                    transform(e.target.value);
```
mz8i: DataEntry allows supplying a valueTransform function, now used only with the postcode field - allows to add custom value processing logic like `value => value.toUpperCase()` for certain fields.

app/src/frontend/building/data-components/like-data-entry.tsx
```diff
@@ -21,29 +29,31 @@ const LikeDataEntry: React.FunctionComponent<any> = (props) => { // TODO: remove
...
-                onChange={props.onLike}
-            />
-            <label htmlFor="like" className="form-check-label">
+            <label className="form-check-label">
+                <input className="form-check-input" type="checkbox"
+                    name="like"
+                    checked={!!props.userLike}
+                    disabled={props.mode === 'view'}
+                    onChange={e => props.onLike(e.target.checked)}
+                />
```
mz8i: the for/htmlFor attribute works correctly only when the referenced id is unique in the scope of the whole page (which was not the case) nesting input inside label is more reliable

mz8i: this is to make the label part of the click area for the checkbox (clicking the label toggles the checkbox) - which is advisable in general but especially on mobile

app/src/frontend/building/data-components/numeric-data-entry.tsx
```diff
@@ -24,7 +34,12 @@ const NumericDataEntry: React.FunctionComponent<any> = (props) => { // TODO: rem
                 min={props.min || 0}
                 disabled={props.mode === 'view' || props.disabled}
                 placeholder={props.placeholder}
-                onChange={props.onChange}
+                onChange={e => 
+                    props.onChange(
+                        props.slug,
+                        e.target.value === '' ? null : parseFloat(e.target.value)
+                    )
+                }
```
mz8i: Parse numbers already inside NumericDataEntry and pass only the resulting value higher up, rather than the whole event. This is to enable correctly detecting changes. Otherwise, if the higher level form component deals with raw string data, initially the field might have a numeric value but after an edit it becomes a string, so detecting edits in the UI is more difficult. The NumericDataEntry, and other types of input components, are specific to the value type they represent, so the parsing of that type is embedded inside the component

app/src/frontend/building/data-container.tsx
```diff
@@ -14,39 +41,55 @@ import InfoBox from '../components/info-box';
  *
  * @param WrappedComponent
  */
-const withCopyEdit = (WrappedComponent) => {
-    return class extends React.Component<any, any> { // TODO: add proper types
+const withCopyEdit = (WrappedComponent: React.ComponentType<CategoryViewProps>) => {
```
mz8i: Through the interface `CategoryViewProps` and the `React.ComponentType` generic type, we provide type checking for the uses of `withCopyEdit`, and better type checking for uses of `WrappedComponent` inside the body of the HOC

app/src/frontend/building/data-container.tsx
```diff
@@ -14,39 +41,55 @@ import InfoBox from '../components/info-box';
...
+    return class DataContainer extends React.Component<DataContainerProps, DataContainerState> {
```
mz8i: Make the local class named, so that the `DataContainer` class name displays e.g. in React Developer Tools in the browser (previously the node in the component tree displayed as an anonymous name with a name like `_a`)

app/src/frontend/building/data-container.tsx
```diff
@@ -14,39 +41,55 @@ import InfoBox from '../components/info-box';
...
+                    keys_to_copy: {},
+                    buildingEdits: {},
+                    currentBuildingId: newBuildingId,
+                    currentBuildingRevisionId: newBuildingRevisionId
+                };
+            }
+
+            return null;
+        }
+
```
mz8i: The static method `getDerivedStateFromProps` is the improved mechanism for making changes to state based on changes to props, and is safer than the deprecated methods like `componentWillReceiveProps`
The state reflects the currently selected building and its revision. If a new building gets selected, or the current building is updated, the edit- and copy-related state is cleared

app/src/frontend/building/data-container.tsx
```diff
@@ -62,7 +105,7 @@ const withCopyEdit = (WrappedComponent) => {
          * @param {string} key
          */
         toggleCopyAttribute(key: string) {
-            const keys = this.state.keys_to_copy;
+            const keys = {...this.state.keys_to_copy};
```
mz8i: copy with object spread, so as not to mutate state

app/src/frontend/building/data-container.tsx
```diff
@@ -73,45 +116,34 @@ const withCopyEdit = (WrappedComponent) => {
...
             });
         }
 
-        /**
-         * Handle changes on typical inputs
-         * - e.g. input[type=text], radio, select, textare
-         *
-         * @param {*} event
-         */
-        handleChange(event) {
```
mz8i: no need for event-processing because input components provide the already parsed value for field (also deleted handleCheck for same reason)

app/src/frontend/building/data-containers/location.tsx
```diff
@@ -96,7 +98,7 @@ const LocationView = (props) => (
             mode={props.mode}
             copy={props.copy}
             step={0.0001}
-            placeholder={51}
+            placeholder="51"
```
mz8i: The placeholder isn't really related to the type of the input value, it can any text to display in the input box so makes more sense to make it a string

app/src/frontend/building/sidebar.css
```diff
@@ -34,6 +34,14 @@
...
     color: #222;
     padding: 0.75rem 0.25rem 0.5rem 0;
+    z-index: 1000;
+}
+
+@media (min-width: 768px) {
+    .section-header {
+        position: sticky;
+        top: 0;
+    }
```
mz8i: On larger viewports where space is not an issue, making the header sticky makes navigating between categories easier

app/src/frontend/helpers.ts
```diff
@@ -36,4 +36,16 @@ function sanitiseURL(string){
...
+    for (const [key, value] of Object.entries(objB)) {
+        if (objA[key] !== value) {
+            reverse[key] = objA[key];
+            forward[key] = value;
+        }
+    }
+    return [forward, reverse];
+}
+
+export { sanitiseURL, compareObjects }
```
mz8i: this was adapted from the comparison function used for calculating edit history records

app/src/frontend/building/data-containers/category-view-props.ts
```diff
@@ -0,0 +1,21 @@
+interface CopyProps {
+    copying: boolean;
+    toggleCopying: () => void;
+    toggleCopyAttribute: (key: string) => void;
+    copyingKey: (key: string) => boolean;
+}
+
+interface CategoryViewProps {
+    intro: string;
+    building: any; // TODO: add Building type with all fields
```
mz8i: even though there is a new `Building` type, in order for this to be used in the category views, it would need to have all attribute fields in the type. This would then require adding the attributes to the building type whenever a new attribute is added. We need to think if this is the best way to go, or if there would be one that would require less modifications in code to add a new category.
The current state is easier to code but it's at the cost of not having stricter type checking.

PR #479, opened 2019-10-28
Feature 189: Adjust likes colour scale and legend

### LINKED ISSUE #189: increase number of likes on like me key and change strapline
a) an amenity society in islington is going to test the like me section to demonstrate community concern at the potential demolition of a local building.  Can we up the top amount to 100? and then do <5, 5-19 20-49 50-99 and > 100 perhaps?  
b)  Also Issue 122 still occurring
c)  Also can we have the hint sentence in a box below the  'I like this building' tick box ( i.e. where the - this takes time to colour sign is)?  So you're being asked explicitly 'Do you like this building and think it contributes to the city?'
d) Ideally you would be able to see colour change every like added - need way of thanking/motivating if intervals

mz8i:
Resolves #189.
New scale:
```
>= 100
70-99
40-69
15-39
5-14
1-4
```
Colour progression left the same as previously

### DIFF
--- a/app/map_styles/polygon.xml
+++ b/app/map_styles/polygon.xml
@@ -222,27 +222,27 @@
     </Style>
     <Style name="likes">
         <Rule>
-            <Filter>[likes] &gt;= 10</Filter>
+            <Filter>[likes] &gt;= 100</Filter>
             <PolygonSymbolizer fill="#bd0026" />
         </Rule>
         <Rule>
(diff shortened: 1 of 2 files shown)


mz8i:
The updated bins and colour scale:
image
I left the upper scale as-is and added one lighter colour for the `1` bin. I think the effect is quite good! Will wait for your opinion @polly64

polly64:
perfect thanks!

Issue #456, opened 2019-10-03
TS interfaces for proptypes of Category and Categories


tomalrussell:
see comment: https://github.com/tomalrussell/colouring-london/pull/403#discussion_r318006280

as part of ongoing work to statically type the codebase

mz8i:
As part of this, I'd suggest implementing a mechanism for generating the React propTypes from the TS interfaces and removing the manual propTypes definitions - see https://github.com/milesj/babel-plugin-typescript-to-proptypes

PR #489, opened 2019-11-06
Auto-generate propTypes

mz8i:
This removes the manual propTypes definitions and replaces them with a mechanism for automatic generation based on TypeScript interfaces. The mechanism currently uses the `babel-plugin-typescript-to-proptypes`, which is missing some more advanced features, for if a prop is of a type that is not defined in the present file, it will end up generating a `PropTypes.any` validation. However, the general opinion in the community seems to be that with TypeScript, propTypes become partially obsolete, and it feels that not having to maintain both mechanisms is an improvement.


app/src/frontend/building/categories.tsx
```diff
@@ -117,12 +121,17 @@ const Categories = (props) => (
...
-    mode: PropTypes.string,
-    building_id: PropTypes.number
+interface CategoryProps {
+    mode: 'view' | 'edit' | 'multi-edit';
+    building_id?: number;
+    slug: string;
+    title: string;
+    desc: string;
+    help: string;
+    inactive: boolean;
```
mz8i: resolves #456

PR #498, opened 2019-11-13
Add data accuracy text to welcome screen

mz8i:
Needed to re-style the welcome page a bit, e.g. decrease the font size, so that the increased amount of text fits fully on most screens. If it still doesn't there will be a scroll-bar that might result in the logos and/or "Start colouring" button being hidden at first, which is not ideal.
See screenshot for current look:
image

### DIFF
--- a/app/src/frontend/pages/welcome.css
+++ b/app/src/frontend/pages/welcome.css
@@ -6,12 +6,14 @@
     z-index: 10000;
     top: 0;
     width: 100%;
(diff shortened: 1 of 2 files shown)


mz8i:
@polly64 please advise if this meets the requirements

mz8i:
@polly64 please see screen after text corrections, added link to GitHub:
image

polly64:
great thanks

PR #503, opened 2019-11-14
Global edit history

### LINKED ISSUE #499: Simple screen to see all recent edits
Create a simple screen showing all changes to the database from the last 3 weeks (time span to be adjusted?). This is to be displayed in a similar fashion as the current building edit history screen, but with edits for all buildings displayed in one place, and each position in the list linking to the edit page of the edited building.
This is a temporary solution to facilitate a retrospective review of recent data edits.

mz8i:
Resolves #499 

The screen is currently not linked to from the menu. Access through `colouring.london/history.html`. Currently will always display the last three weeks of changes. To be further developed.

### DIFF
--- a/app/src/frontend/app.tsx
+++ b/app/src/frontend/app.tsx
@@ -9,6 +9,7 @@ import MapApp from './map-app';
 import { Building } from './models/building';
 import { User } from './models/user';
 import AboutPage from './pages/about';
+import ChangesPage from './pages/changes';
 import ContactPage from './pages/contact';
 import ContributorAgreementPage from './pages/contributor-agreement';
 import DataAccuracyPage from './pages/data-accuracy';
@@ -112,6 +113,7 @@ class App extends React.Component<AppProps, AppState> {
                 <Route exact path="/data-accuracy.html" component={DataAccuracyPage} />
                 <Route exact path="/data-extracts.html" component={DataExtracts} />
                 <Route exact path="/contact.html" component={ContactPage} />
+                <Route exact path="/history.html" component={ChangesPage} />
                 <Route exact path={App.mapAppPaths} render={(props) => (
                     <MapApp
                         {...props}
--- a/app/src/frontend/data_fields.ts
+++ b/app/src/frontend/data_fields.ts
@@ -11,6 +11,8 @@ export enum Category {
     Community = 'Community',
     Planning = 'Planning',
     Like = 'Like',
+
+    Unknown = 'Unknown'
 }
 
 export const categories = {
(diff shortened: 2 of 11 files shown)


app/src/frontend/map-app.tsx
```diff
@@ -56,6 +58,7 @@ class MapApp extends React.Component<MapAppProps, MapAppState> {
 
     componentDidMount() {
         this.fetchLatestRevision();
+        this.fetchBuildingData();
```
mz8i: This needs to happen when a building page (e.g. `/view/age/233771`) is navigated to from another place in the app. Previously there were no links to building pages in the app, so the data was either loaded in `frontendRoute.tsx` as part of server-side rendering, or it was loaded upon selecting a building on the map.
Now, with the addition of global edit history screen, the edit records link to building pages, and so there needs to be a mechanism to load the data when a component is mounted but there is no pre-loaded data from SSR.

app/src/frontend/map-app.tsx
```diff
@@ -75,6 +78,52 @@ class MapApp extends React.Component<MapAppProps, MapAppState> {
...
+                        credentials: 'same-origin'
+                    }).then(res => res.json()),
+                    fetch(`/api/buildings/${buildingId}/like.json`, {
+                        method: 'GET',
+                        headers: {
+                            'Content-Type': 'application/json'
+                        }, 
+                        credentials: 'same-origin'
+                    }).then(res => res.json())
+                ]);
```
mz8i: We should probably look into writing some helpers for making JSON API calls (or use a third-party library like Axios) to reduce `fetch`-related repetition across the front-end.

Also, it might be good to create a special endpoint for fetching all relevant data for a building/user pair, to reduce the number of requests to the back-end.

app/src/frontend/app.tsx
```diff
@@ -112,6 +113,7 @@ class App extends React.Component<AppProps, AppState> {
                 <Route exact path="/data-accuracy.html" component={DataAccuracyPage} />
                 <Route exact path="/data-extracts.html" component={DataExtracts} />
                 <Route exact path="/contact.html" component={ContactPage} />
+                <Route exact path="/history.html" component={ChangesPage} />
```
mz8i: The history page is accessible but there is currently no link pointing to it, given its very rough form.

app/src/frontend/data_fields.ts
```diff
@@ -1,18 +1,71 @@
...
     Streetscape = 'Streetscape',
     Team = 'Team',
     Sustainability = 'Sustainability',
     Community = 'Community',
     Planning = 'Planning',
-    Like = 'Like Me!'
+    Like = 'Like',
+
+    Unknown = 'Unknown'
 }
```
mz8i: The string enum values now reflect the name of the key. The human-readable values were moved to the separate `categories` definition object which also contains the slug values

PR #530, opened 2019-12-10
Update npm packages

tomalrussell:
- in particular, bump `serialize-javascript` to ^2.1.1


mz8i:
On a related note, I was thinking it might be good to update Node and Postgres to latest stable version at some point soon? For added security fixes and new features. What do you think @tomalrussell? For Node that would be 12.13.1, and Postgres - 12.1

Issue #496, opened 2019-11-12
Initial user leaderboard


tomalrussell:
Towards #48 - first implementation.

- create API endpoint to query contributions
- count contributions simply as number of entries in edit log
- create a new page with table of top 10 all-time contributors

Rank | Username | Contributions
--- | --- | ---
1 | abc | 1234
2 | zzz | 23
... | ... | ...

- add switch between showing top 10/100 
- add switch between showing contributions in the last 7 days/30 days/all time

martin-dj:
I'll have a go at this issue as I've already created the appropriate SQL query to generate some stats for Duncan's Innovate meeting.

tomalrussell:
Basic leaderboard available at https://colouring.london/leaderboard.html

PR #535, opened 2019-12-17
Implementation of basic leaderboard

martin-dj:
Implementation of leader board as outlined in issue #496.

### DIFF
--- a/app/src/api/services/leaderboard.ts
+++ b/app/src/api/services/leaderboard.ts
@@ -12,7 +12,7 @@ async function getLeaders(number_limit: number, time_limit: number) {
                 AND NOT (users.username = 'colouringlondon')
 		GROUP by users.username
 		ORDER BY number_edits DESC
-		LIMIT $2`, [number_limit, time_limit]
+		LIMIT $2`, [time_limit, number_limit]
 	    );
 			
 	}else{
--- a/app/src/api/services/leaderboard.ts
+++ b/app/src/api/services/leaderboard.ts
@@ -1,0 +1,39 @@
+import db from '../../db';
+
+async function getLeaders(number_limit: number, time_limit: number) {
+    try {
+        if(time_limit > 0){
+	    return await db.manyOrNone(
+		`SELECT count(log_id) as number_edits, username
+		FROM logs, users
+		WHERE logs.user_id=users.user_id
+		AND CURRENT_TIMESTAMP::DATE - log_timestamp::DATE <= $1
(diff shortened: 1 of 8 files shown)


martin-dj:
Submitted as feature branch: feature/initial_leaderboard

Pull request #535

app/src/frontend/pages/leaderboard.tsx
```diff
@@ -0,0 +1,149 @@
...
+    }
+	
+	
+    renderTableData() {
+        return this.state.leaders.map((u, i) => {
+	    const username = u.username;             
+            const number_edits = u.number_edits;
+         
+	    return (
+		<tr key={i}>
```
mz8i: the `key` prop is used by React to determine changes to the component tree. Therefore, it's best to go for something that specifically identifies the unique component - `username` here might be better than position in list.

app/src/frontend/pages/leaderboard.tsx
```diff
@@ -0,0 +1,149 @@
...
+        leaders: Leader[];
+        fetching: boolean;
+
+        //We need to track the state of the radio buttons to ensure their current state is shown correctly when the view is (re)rendered
+        number_limit: number;
+        time_limit: number;
+    }
+
+
+class LeaderBoardPage extends Component<LeaderboardProps, LeaderboardState> {
```
mz8i: let's make the class and interface names consistent in terms of `Leaderboard`/`LeaderBoard` casing (not sure which one is more correct)

app/src/api/services/leaderboard.ts
```diff
@@ -0,0 +1,39 @@
+import db from '../../db';
+
+async function getLeaders(number_limit: number, time_limit: number) {
```
mz8i: as far as I understand, this query will be run on the logs table every time the leaderboard page is displayed. Just wondering about the impact on the server's performance? Seeing that the logs table is / will become potentially quite huge with time...

martin-dj: I tested the implementation in my dev environment, against a database dump from the production server (dumped circa September) without issues, despite the limited resources of my virtual dev environment. 

So I would suggest merging it and seeing how it performs.

app/src/frontend/pages/leaderboard.tsx
martin-dj: Fixed

app/src/frontend/pages/leaderboard.tsx
martin-dj: Fixed

PR #546, opened 2020-01-09
Feature 513: activate land use - make editable [WIP]

mz8i:
Main changes from the UI perspective:
-) multi-data-entry lists have been changed a bit. Previously, the plus button would cause adding a new empty field to the list. Now, an additional input box is always displayed, where the user can input their new item and only then click the plus button next to it, in order to add the value to the list.

Original multi-data-entry:
image
image

New multi-data-entry:
image
image

I personally feel this is a bit more consistent, e.g. that the default value is actually a `[]` and not `[null]`. I think the styling (spacing, typography) and labelling of the buttons might still need some adjustments to make the intention clear so please comment on that.

-) an autofill dropdown is now available on text fields that will contact the backend for a list of options. The backend currently uses the `pg_trgm` extension for fuzzy searching for options. This is because of the complexity of the landuse class field, however just using the '%' string similarity operator doesn't work quite well, it only works when the length of the query is close to the length of the class name - this still needs more work and I'd welcome suggestions on how we can improve that.
autofill_screenshot_1
autofill_screenshot_2
autofill_screenshot_3

-) editing of the Current Landuse Group field is disabled when there are any values in the Current Landuse Class field
image
image

-) one aspect that hasn't yet been tackled is the backend for validating edits and deriving higher-level values from lower-level ones. This has been partially implemented before the Paris conference but needs to be corrected and extended.


mz8i:
On the search results, it seems to me that at least for the landuse class field, it would be better if the autofill backend used Postgres full-text search, but I need to investigate into that more

mz8i:
Switched Land Use Class autofill search to use Postgres full-text search rather than `pg_trgm` and I think the suggestion results are better in general. For now the change is just in the query. We might want to add a migration with a full-text search index if we like the behaviour - can you test locally @tomalrussell @dominijk ?

PR #562, opened 2020-02-03
Feature: edit history paging

### LINKED ISSUE #561: Global edit history screen with pagination
Following on from #499 - the edit history page needs to be re-worked to include pagination, so that the recent changes can be browsed without downloading the huge history of edits in one go.

mz8i:
Resolves #561 - adds pagination to edit history endpoint and web page.
Note that this is pagination of a feed of events, which makes it similar to the likes of a Twitter feed (as opposed to pagination of a fixed set of records, where the pages are more or less fixed).
The main navigation mechanism allows for going through the history of changes by loading older or more recent changes than the ones currently shown. There is no indication for the total number of pages.

Some recently discussed mechanisms that were evaluated and implemented here are: separation of data access layer / service layer in API, mocking and unit testing, better error handling in the API, better parameter parsing in the API.


app/src/api/api.ts
```diff
@@ -93,14 +95,30 @@ server.get('/search', function (req, res) {
...
+                errorMessage = err.message;
+            }
+
+            res.status(400).send({ error: errorMessage });
+        } else if(err instanceof DatabaseError){
+            res.status(500).send({ error: 'Database error' });
+        } else {
+            res.status(500).send({ error: 'Server error' });
+        }
+
```
mz8i: Add global handling for typical error situations - this way the individual controllers don't need to all include the code to handle database errors, or errors related to parsing a HTTP request parameter.

app/src/api/controllers/editHistoryController.ts
```diff
@@ -1,17 +1,32 @@
...
-        });
+        const result = await editHistoryService.getGlobalEditHistory(beforeId, afterId, count);
+        res.send(result);
     } catch(error) {
-        console.error(error);
-        res.send({ error: 'Database error' });
+        if(error instanceof ArgumentError && error.argumentName === 'count') {
+            const apiErr = new ApiParamError(error.message, 'count');
+            throw apiErr;
+        }
```
mz8i: Map between `ArgumentError` which doesn't know anything about the service being called inside an API controller, to an API-focused `ApiParamError`. This is useful for determining a situation where the `count` parameter is correctly parsed as an integer, but it doesn't pass some constraint checks defined inside the business-logic-oriented edit history service. The mapping to an `ApiParamError` is necessary for the global error handler to send an HTTP response with the correct (user-error) code (400) rather than generic server error code (500)

app/src/api/dataAccess/__mocks__/editHistory.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Create an object mocking all method of editHistory dataAccess
+ * The type is set to reflect the type of that module, with added methods
+ * used when testing
+ */
+const mockEditHistory =
+    jest.genMockFromModule('../editHistory') as typeof import('../editHistory') & {
+        __setHistory: (mockHistoryData: EditHistoryEntry[]) => void
+    };
```
mz8i: Setting the type to that of the original mocked module, and extending it with a local type containing the `__setHistory` method, allows for checking that the method mocks in this file align with the definitions in the original module, while allowing to use special methods needed to set up the mock during testing.

app/src/api/dataAccess/__mocks__/editHistory.ts
```diff
@@ -0,0 +1,62 @@
...
+    const historyAfterId = await mockEditHistory.getHistoryAfterId(id, 1);
+    return historyAfterId[historyAfterId.length - 1]?.revision_id;
+};
+
+mockEditHistory.getIdOlderThan = async function(id: string): Promise<string> {
+    const historyBeforeId = await mockEditHistory.getHistoryBeforeId(id, 1);
+    return historyBeforeId[0]?.revision_id;
+};
+
+const {
```
mz8i: The `mockEditHistory` object needs to be destructured into individual variables, in order to mimic the export structure of the original `editHistory` module. This could be avoided if every function was defined from the start as a local variable, but we'd lose the benefits of type-checking that the mock matches the original (which cannot be achieved without having `mockEditHistory` as a single object).

app/src/api/dataAccess/editHistory.ts
```diff
@@ -0,0 +1,88 @@
...
+                ${baseQuery}
+                WHERE log_id > $1
+                ORDER BY revision_id ASC
+                LIMIT $2
+            ) AS result_asc ORDER BY revision_id DESC`,
+            [id, count]
+        );
+    } catch(err) {
+        throw new DatabaseError(err);
+    }
```
mz8i: This and other cases of using `DatabaseError` is a test of handling database-related errors in a more consistent way. Wrapping the errors in our custom error class currently serves the only purpose of being able to distinguish, in the global error handler, between a database-related error and a different, generic server error (`pg-promise` does not have a single base class for errors as some are simply piped through from the driver - `node-postgres`)
In the case of other data access layer classes, the error handler might do some more complex error checking, for example to check if the error is actually due to the value of the arguments, or due to a connection problem.

app/src/api/parameters.ts
```diff
@@ -0,0 +1,44 @@
...
+    try {
+        return processingFn(stringValue);
+    } catch(error) {
+        if(error instanceof ApiParamError) {
+            error.paramName = paramName;
+        }
+        
+        throw error;
+    }
+}
```
mz8i: The `processParam` function handles the parsing of a request parameter. This is in lieu of using an actual library/middleware for parsing parameters.
The function needs to know the name of the parameter, so that it can assemble an appropriate error message if there is any problem with the parsing. Parameters can be specified as required. The actual parsing function can be any function that processes the value to a type different than string, or that validates the format of the string parameter (examples below: parsing an ID, validating against regex). The individual parsing functions should handle undefined parameters gracefully.

Issue #543, opened 2020-01-09
UI: Simple fixes - colours, titles, legend


polly64:
@mz8i  could u possibly have a look at these if they are quick fixes for you? many thanks
- [x] 1. Adjust main category titles ready for Tom redesign - as last comment #490
- [x] 2. Adjust these main also for titles which appear when main category buttons are clicked #541
- [x] 3. Adjust single colour on land use map #541
- [x] 4. Adjust building age colours #525
- [x] 5. ~~Advise/help Dom  adding conservation area boundaries to new listed building map~~ (replaced with CA outline for each building separately) #242
- [x] 5. Remove vacant and derelict from Land use key #529
- [x] 6. Add small colour logo to each map key #236
- [x] 7. Add simple text list info under new main Dynamics title ( for info due to be collected) as 

follows: 
- Under threat of demolition (partial/complete? 
- Demolition permit no. issued, Whole building demolitions for current year, Whole building demolitions since 2000, Pairs of construction and demolition dates for previous buildings built on any part of the site. @polly64 can you confirm the edit I've just made is correct (the bullet pointing)
thank you!

- [x] 8. Change `Size & Form` category to visualise to height as Slack [thread](https://colouringlondon.slack.com/archives/CM3PJTHQV/p1578647874013400) 
- [x] 9. Change `Planning` category to visualise by listed buildings `Listing grade` which links to #538

dominijk:
@polly64 Did you also want to you want to add these two as well?

1. Change `Size & Form` category to visualise to height as Slack [thread](https://colouringlondon.slack.com/archives/CM3PJTHQV/p1578647874013400) 
2. Change `Planning` category to visualise by listed buildings `Listing grade` which links to #538

polly64:
yes please

mz8i:
@polly64 number 2 on the above list links to the same issue as number 3, I think this is a typo - could you please check which issue it was supposed to link?

mz8i:
ticked all items that are already in review.

polly64:
@mz8i shall i keep adding small fixes to this page only?

mz8i:
@polly64 I have created a new label for issues: `small` - please create new issues for each of these tweaks and label them as `small` (using the Labels menu on the right) - then if someone goes to https://github.com/tomalrussell/colouring-london/labels/small they will be able to see all issues labelled as `small`. It will be easier to manage the completion of these tweaks if they are separate.

Issue #529, opened 2019-12-07
Land use dataset remove empty and vacant category


polly64:
Can we remove for the moment off key, @dominijk and @polly64 to discuss collection of this field

dominijk:
@mz8i I think we can adjust the map legend simply enough? Logically we'd then want to to omit the option on the input, is that work more involved. My feelings are we don't want to edit the database design at this stage.

Issue #391, opened 2019-08-15
Revise contributor agreement and/or introduce code of conduct


polly64:
Do we have anything in the sign up stating that malicious uploads will result in the deletion of accounts and that is at our discretion?

tomalrussell:
Not at present.

Our contributor agreement focusses on data contributions: https://colouring.london/contributor-agreement.html

On the discuss forum we have some guidelines: https://discuss.colouring.london/faq

Here are a couple of options suggested by GitHub's guidelines for project code of conduct:
- recommended for projects of any size: https://www.contributor-covenant.org/version/1/4/code-of-conduct
- recommended for larger communities and events:  http://stumptownsyndicate.org/code-of-conduct/

The first should be easy to put in place; the second would require a bit more organisational structure first.

Issue #264, opened 2019-05-04
Menu redesign


polly64:
[Edited 4/10/19 to suggest that this be taken as part of a main menu redesign]

Main menu should link to:
- main features
  - View/Edit Maps
  - Downloads
  - Showcase
  - Leaderboards
  - Discuss
  - Log In, Sign Up // Account
- major info pages
  - About
  - Data Categories
  - Who's Involved
  - Data Ethics
- categories 
  - all 12 boxes
- minor info pages
  - Contact
  - Privacy Policy
  - Contributor Agreement
  - Code of Conduct
  - Data Accuracy Agreement
  - Ordnance Survey terms of UPRN usage

---
Below is the original draft for landing page Menu done with louis for the exhibition , would be good to now introduce this along with new draft logo- so we get clear branding message across 
Would be good to land on this page after main landing page  (and/or straight after sign up) so can choose to go straight to View or edit and see we will have Download and showcase but with hover showing they're coming soon.  I think it's important as it allows us to show whole platform strategy scope in one go - and to advertise major forthcoming features. 

Attach image - Louis can send logo
[190503_Colouring London PAGE 2.pdf](https://github.com/tomalrussell/colouring-london/files/3144472/190503_Colouring.London.PAGE.2.pdf)

dominijk:
@polly64 Does this come after this page? 
<img width="1064" alt="Screen Shot 2019-07-01 at 18 08 56" src="https://user-images.githubusercontent.com/6041913/60454276-54b22a80-9c2b-11e9-926b-ca8add4b3800.png">

tomalrussell:
Closed by #564

PR #564, opened 2020-02-03
Feature: menu sidebar

### LINKED ISSUE #490: revised view map & titles  interface visual
revised draft
category info margin widened to accommodate clickable category/logo. Avoids backspace.  needs to be wide enough to be legible. Top bar moved to menu box- n(eeds icon and text description) following tom thought we could gain more space losing it completely.  key made narrower. Category highlights when clicked and categories open below.  Category box stays in place so you can easily add more data for building. logo shrunk on key. Do add comments

tomalrussell:
Aiming to close #490 and #264 with a redesign/layout of the main menu.

Small fixes (which don't touch the map styles/tiles) from #543 :
- 1 Adjust main category titles ready for Tom redesign - as last comment #490
- 5 Remove vacant and derelict from Land use key #529
- 6 Add small colour logo to each map key #236
- 7 Add simple text list info under new main Dynamics title ( for info due to be collected)

Also (!) adds draft code of conduct as suggested in #391


tomalrussell:
Snags:
- [ ] awkward layout on mid-sized screens (700-1000px) - try narrower sidebar
- [ ] more spacing between category squares
- [ ] white background/whitespace instead of coloured bar "Age"/etc. when viewing/editing a particular building

mz8i:
What I like:
- The general split of the menu into sections (wondering if Discuss should be moved to second section so that we exclude any links to external websites from the first section?).
- On desktop, having the category links in squares makes it quicker to switch between different views, which is a good change.
- On mobile, having the menu hide to the side allows to see much more of the map than before, which is great.

A couple doubts:
- the squares take up quite a lot of valuable space actually - both on desktop and mobile (at the cost of displaying the map and the data in the sidebar). This gets quite extreme on small mobile screens.
- on mobile, viewing data from different categories for the same building is slightly simplified (single tap rather than two to switch between categories) but at the cost of making it much harder to see data from many buildings for the same category. The user needs to collapse the sidebar, select a new building, expand the sidebar etc. Perhaps inspecting all data for a single building is the case we want to optimise for but as it stands, it's impossible to view both the data and the map at the same time.

Nitpicks (not fundamental but probably best to be resolved/discussed ahead of releasing):
- [ ] not sure if the two sets of squares (in the menu / in the sidebar) are needed? Might lead to confusion, e.g. the sidebar category links lead to `view` or `edit` depending on the mode, while the menu ones always lead to `view` etc.
- [ ] the sidebar squares have some problems with size/alignment and anti-aliasing (see picture, screenshot taken on Android but same happens on desktop Firefox)
- [ ] I've only seen this happen on Firefox, but when clicking on one of the sidebar categories, the appearing scrollbar causes the squares to overflow and the grid becomes a 4x3 instead of 3x4, taking even more vertical space
- [ ] the first two category squares in the last row (planning, sustainability) have swapped colors when comparing the logo (blue, green) to sidebar menu (green, blue)
- [ ] when coming onto the landing page, the empty white sidebar looks a bit out of place, gives the feeling of something unfinished. Maybe we could show the categories immediately?
- [ ] on mobile, there is no map zoom. It's possible to pinch-zoom on mobile, but that requires using two hands instead of one and unfortunately isn't very fluid with Leaflet on mobile web. I'd suggest keeping the zoom and search controls on all screen sizes

Issue #538, opened 2020-01-03
PLANNING & DESIGNATED BUILDINGS


polly64:
@mz8i To do list - PH still to complete

1- [ ] Tidy up by Concertinering all duplicate options i.e. don't show more details for GHLER, listing etc  until 'Yes' option has been clicked

Check these with PH colouring-cities/colouring-core#574

---- 

Update to planning category to switch map visualisation from conservation areas (as current) to listed and protected buildings:

Linked issues colouring-cities/colouring-core#300 colouring-cities/colouring-core#232 colouring-cities/colouring-core#160 

- Switch from single field categorisation to multi-field categorisation
  - `planning_in_conservation_area` - If true colour as below below else don't render
  - `planning_list_cat` - visualised if `Listed building` null (use grade as below), else if `Scheduled Monument` use colour below - No data added see [issue](https://github.com/colouring-london/colouring_london_analysis/issues/16)
  - `planning_list_grade` -   visualised as below colours/categories
  - `planning_in_local_list` - If true colour as below else don't render - Data added see [issue](https://github.com/colouring-london/colouring_london_analysis/issues/15) 
- On category load expand to show `Listing and protections` ? Team discus
- Lock front end entry for `National Heritage List for England list id`, `National Heritage List for England list type
` , `Listing grade` and `Is listed on the National Heritage List for England?` 
- ~~Update basemap  to include conservation areas see colouring-cities/colouring-core#242~~ 
- Add verification box to front end (disabled)
- Add disclaimer to legend and below sub title that the only authoritative source of listed buildings data is HE and not to rely on CL data for list status or extent of listing
Disclaimer text: "_All data relating to designated buildings should be checked on the National Heritage List for England or local authority websites where used for planning or development purposes_"
- Add new legend for visualisation as below colours 
Prepopulate NHLE links colouring-cities/colouring-core#354
Discussion on boundary lines colouring-cities/colouring-core#242
Mapping planning applications -post June 2019 colouring-cities/colouring-core#211

Colour key  references for Grade 1, Grade 11*, Grade 2 and locally listed buildings

name | rgb | hex
--- | --- | ---
Grade 1 (Red)  | `rgb(199,46,8)` | `#c72e08`
Grade 11* (Gold)  | `rgb(230,152,0)` | `#e69800`
Grade 11  (Grey/Blue)  | `rgb(158,187,201)` | `#9ebbc9`
[...]

Issue #525, opened 2019-11-29
colour keys for age maps


polly64:
New colours for 2 x categories CL age map as some needed enriching for better user 'thank you' and also we are likely prepopulate residential categories for semi detached and victorian (me working with Flora who's at turing). I'm also merging 3 categories as unnecessarily detailed

So 1880-1899 changes to  
C=20
M=72
Y=44
K=21
or R = 150, G = 18, B=89

1860-1879 changes to
C=71
M=67
Y=35
K=0
or R=72, G=84, B=166

Merge 1800-1819 with 1820-1839 same colour as Tarn's original colour for 1780-1800

Merge all categories from 1700-1799 and make same colour and rename 1700-1799. Use Tarn;s original 1760-1780 colour

Merge all categories from <1699 and make the colour of this category white

polly64:
hi Maciek can u also adjust the dates - sorry they should  read 1920- 1929, 1930-1939 etc not 1920-1930, 1930-1940

mz8i:
Hi @polly64 by Tarn's original colour for 1760-80 do you mean the current colour for that range?

mz8i:
@polly64 I have implemented the changes, but need to consult a few things before releasing

1. the first new colour you gave, the CMYK colour doesn't match the RGB colour (see below) - which one to use, the RGB definition or CMYK definition? It's a small but visible difference.
`rgb(161, 56, 113)` ( = cmyk(20%, 72%, 44%, 21%))
`rgb(150, 18, 89)`

2. that's how the legend looks after the colour scale changes (no adjustment yet to 1920->1919 etc):
![image](https://user-images.githubusercontent.com/36160844/74042829-2d40ee80-49c0-11ea-8e3f-c85eeae31b3e.png)
Is this the desired progression? At first sight it looks a bit like 1840-1860 and 1860-1880 were meant to be swapped?
3. I think in another discussion @polly64 you were saying we shouldn't use white colour because it is difficult to see on the light style background (see St Paul's Cathedral below). Do we want to stick to white as the oldest colour?
![image](https://user-images.githubusercontent.com/36160844/74043122-bb1cd980-49c0-11ea-98d0-4e9734bc2a27.png)


@tomalrussell I assume that after the change, we need to purge the tile cache for the whole category? We might want a script to re-cache a certain region, e.g. overnight - for changes like this and for new categories (before releasing)

mz8i:
[...]

Issue #160, opened 2018-12-15
Listed buildings and conservation area data


urbandataanalysts:
@Polly64 Two questions for you. Along with the fields on the control page do we also want to add the date the building was listed? Similarly conservation areas also have a delegation date when they came into force which we can include if you'd like?

For boroughs who haven't released their conservation area data this won't be that visually obvious we'd need to flag the buildings in the borough to not colour or add overlay. One to discuss. 

@tomalrussell I'll also include the listed building list entry number, this can then be used to generate the URL to the entry as below;

List entry number : **1026593**
URL is this plus standard stub
https://historicengland.org.uk/listing/the-list/list-entry/1026593

For the energy project we are doing this in the js, what would you like to do here? I can generate the links in SQL and have them in db or we can generate them OTF from the number

polly64:
1) Yes please, designation dates for both will be very useful for future analysis.
let talk re encouraging borough uploads re CA data. 
2) As soon as CA data are  visualised we can write to the missing at in a single email.
3) Can we put constraints on URL entry format ie can you say it always has to have the historic england prefix and then include a numeric ID as these will always be the same?

tomalrussell:
For anything where a URL can be generated from ID or similar, just the ID would be ideal - generating links on the fly should be safer and simpler.

polly64:
Dom could you have a think how data sources would work in the Planning system?  We need to try and get this on as soon as we can.  Tom presumably if someone edits something then the source box will clear too for them to put new source in?

urbandataanalysts:
@polly64 I'm not sure I've fully understood the query, is this to think what people might do with the data? I've lost track on a few items below;

### What level of editing are we allowing on statutory designated items (Listed buildings / Conservation areas)? 
I think when we last spoke we discussed allowing people to flag if they thought it was incorrect perhaps they could also have a free text box in which to suggest source or reason but this data is hidden. I think the risk of someone maliciously editing is lower than an accidental deletion/edit. 

### Data source versus input user
[...]

PR #573, opened 2020-02-09
Feature 538: planning update

mz8i:
Work towards #538.

Still missing: 
- disclaimer about checking listed buildings status in official sources
- verification box (need clarification?)


dominijk:
Looks good, I can see that I only loaded the listed buildings data into production so that explains why there are so few on staging but good to see the locally listed being picked up. I think the grey colour for conservation areas is too subtle on the lighter background map to really pop out. 

One that would be good to wrap into this update would be the auto generation of the Listed building URL to populate the front end as discussed on #160 the list entry is stored in `buildings.planning_list_id` and is `int4` 

@mz8i On the disclaimer do you need more on the wording for this. I can see the dscussion on wording from [Slack](https://colouringlondon.slack.com/archives/CM3PJTHQV/p1578658770028900) thread is superseded by this wording from issue #538   @polly64 can you review and confirm, also I've lost track of whether demolitions is remaining in this category or moving to dynamics?

> Add disclaimer to legend and below sub title that the only authoritative source of listed buildings data is HE and not to rely on CL data for list status or extent of listing
> Disclaimer text: "All data relating to designated buildings should be checked on the National Heritage List for England or local authority websites where used for planning or development purposes"

mz8i:
@dominijk The disclaimer text in the issue above was copied by me from the Slack thread so that we don't have links to external communication but rather all information in this issue. If I missed some later discussion, can you please amend the text here to reflect the final version?

polly64:
@mz8i @dominijk  yes could you use the disclaimer above and then i will show Historic England at meeting on wed morning. At the top of the column if ok?  if we can get all of them loaded by then would be great. @mz8i if  can you possibly swap the white with border to the scheduled monuments to represent conservation so I can have a look and then make the scheduled monuments great with a white border would be great

mz8i:
Now also includes work towards #525 to make it easier to test both on staging

polly64:
@mz8i Re query about demolition - 
Could u add a new separate section below the top 'planning portal link ' box
So it reads 
1. 'Planning Status'
2. 'Designation and Protection' (not 'listing and protection')
Then just keep everything already under the 'Designation & Protection' section as it is

When u open up 'Planning Status' section could u make 4 new boxes:
1. 'Is a planning application live for this site?' (make this inactive) 
2. 'Is demolition proposed?'  (make this live) as a tick box
3. 'Has this application recently been been approved/refused?'  (make this inactive)
4. 'Has the work been carried out? (make this inactive)
Then add a note below this
'For historical planning applications see Planning Portal link' 

Ideally put hover over the first two inactive categories saying 'future proposed live link with Planning portal' but not essential
Move the rest of the demolition section to Dynamics

mz8i:
@polly64 can we move this into a dedicated issue?

mz8i:
@polly64 I will change the names but will keep all the fields in Planning Status inactive - the planned demolition field currently doesn't exist in the database, and I'd like to prevent the size of this piece of work from getting out of hand - can we work on adding / activating that field as part of a separate issue?

mz8i:
@polly64 @dominijk please see staging for latest update of planning and age map. As I have modified the colour scale in both cases (fundamentally in the case of planning), this is just an experiment, and please feel free to say we should revert to the previous options and try tweaking them further.

For Age, I made some further tweaks, while re-using as many of Tarn's colours as possible  (added only one colour - the light blue in the middle).

For Planning, I tried to make it work with the requested colours but couldn't manage to make it look good with small tweaks, so decided to try a completely different approach. I based it on the following reasoning:
- white and gray are difficult to see on either the dark or light background map, so are best avoided
- conservation areas take most space so the colour should be least "aggressive" - chose a light turquoise
- listed buildings are most important (so should contrast with the background of conservation areas - red is good) but grades I, II* and II are levels of the same classification - kept red for Grade I and adopted lighted shades of the same colour for the lower grades
- scheduled monuments also come from the heritage list, but are different than listed buildings - chose yellow
- locally listed buildings are yet another class of objects - chose violet

`#c72e08` Grade I
`#e75b42` Grade II*
`#ffbea1` Grade II
`#ffe59b` Scheduled Monument
`#858ed4` Locally Listed
`#95beba` In Conservation Area

Again this is just a test so all comments welcome. Please let me know if we want to tweak some of these colours or revert to tweaking the previous scheme

polly64:
@mz8i @dominijk hi I can only see one or two listed buildings on staging are they up? I will talk tarn about colours - I just need to get her a screen shot when I can see everything on there- do u think it's just they're not loading on mine or are they not up yet?

dominijk:
@polly64 As earlier comment listed buildings are loaded into production but not staging. So that's why you aren't seeing many. @mz8i and Polly, I think this looks good, some comments below. I agree with raising new issue with order of the boxes on front end. Let's keep this just about colours and what is being coloured;
- Title: in place of planning Statutory protections
- Legend: Move conservation areas higher up as it's the most used
- I'm not sure we will end up with scheduled monuments, these seem to often be geographic boundaries that underly listed buildings

mz8i:
@dominijk any chance we could load the same data to staging? it would help massively in terms of evaluating the colours

polly64:
@mz8i @dominijk also even  i could get a screengrab showing even a small section with all categories coloured in draft and the disclaimer on the key as it is would be great for my grant meeting tomorrow

mz8i:
@polly64 I might copy some data directly from production database to staging so that we have something by tomorrow

polly64:
@mz8i that would be completely brilliant.

Issue #568, opened 2020-02-05
UI: Activate construction category & land use 


dominijk:
References #359 #361 

Below steps are taken from the [docs](https://github.com/tomalrussell/colouring-london/blob/master/docs/adding-new-fields.md) for how we add a new category.  Map will be coloured by primary material.  

- [ ] Set up database, adding primary material ,roof material and secondary materials using material options as #539 with constraint (entry on website is drop down)
- [ ] Set up API elements
- [ ] Set up front end, entry is drop down for primary material ,roof material and secondary materials 
- [ ] Set up data extracts
- [ ] Setup tile server - colours for map tiles as #539 
- [ ] Set up legend 
- [ ] Add bulk data set see [issue](https://github.com/colouring-london/colouring_london_analysis/issues/18)

polly64:
@tomalrussell Interface really starting to look great. Will put all comments below.

Main interface page
1. On my computer the  4x3 logo is turning into a 3x4- can we do something to make it stay the same as menu  one
2) I agree if menu can collapse so u can see map before would be good but needs to be really clear as to where it's gone.  I really like way menu says menu and has menu logo. Everything just needs to be super clear- I think though that this should go in to next to do list and we shouldn't wait for this to release.
2. When u click on a category a coloured category title comes up.  I think this clashes a bit with the logo and it would be better to  a) keep a highlight around the selected  category on the 12box logo (say in thick light grey or possibly turquoise as on menu) and b) have a white rather than a coloured background for the category title with a thin border (same as grey border on box below). Also have   this title box align with left hand side of 12 box logo/categories  and match the width and alignment of data entry boxes below.  

Materials
1. Can we upload melda's listed buildings as also will be useful re showing for grant (esp as we have have to rethink approach to designation)
2. Can we swap the colour allocated for' wood' for the colour allocated for 'other natural material'? as the wood colour i realise is going to be quite dark and unsatisfying to colour in- the brick , steel and RC colours are really nice to use.
3. I really like the way the copy tool is so clearly marked now, My only thought is i think there is  one too many steps and it's not quite intuitive yet.  So it feels fine when i click' copy' and have to tick the copy box but then I really  i need a pop up hint or something to remind me to click copy selected. Could we add this?

Land use
1. Can we swap the ' dwelling' blue colour (as this is too dark to be satisfying to colour in and likely comprise over 90% of entries) with the turquoise representing 'community services'
[...]

Issue #539, opened 2020-01-03
Colour key references: Primary materials and roof materials


polly64:
Colour key references for both Primary materials and roof materials as follows: if additional categories are required we can reassess scheme and add a colour.

name | rgb | hex
--- | --- | ---
Brick/Clay (Pink)   | `rgb(252, 168, 157)` |  `#fca89d`
Slate (dusty pink)  | `rgb(196, 138, 133)` |  `#c48a85`
Steel/Metal (Forest green)  | `rgb(92, 137, 112)` |  `#5c8970`
Timber (Brown)  | `rgb(150, 97, 59)` |  `#96613b`
Stone (kharki green)    | `rgb(181, 168, 89)` |  `#b5a859`
Glass (cream)   | `rgb(255, 255, 227)` |  `#ffffe3`
Concrete (duckegg blue) | `rgb(190, 255, 232)` |  `#beffe8`
Natural (including green roofs/thatch)  | `rgb(245, 217, 107)` |  `#f5d96b`
Ashphalt or other (DarkBrown/Purple)    | `rgb(77, 0, 0)` |  `#4d0000`

dominijk:
Can we have the colours as a hex reference as this is what we use in the code, this is a nice website for converting and picking colours https://htmlcolorcodes.com/color-picker/

polly64:
Hi Dom i have to use the arcgis system as I test them on the maps but think they have RGB, CMYK and HEX on it


Polly Hudson

Senior Research Fellow

Director - Colouring London https://colouringlondon.org/

Bartlett Centre for Advanced Spatial Analysis

University College London

90 Tottenham Court Road W1T 4TJ



________________________________

tomalrussell:
Just updated - you can search directly for colours on https://www.colorhexa.com/

dominijk:
@polly64 , @MeldaS and I reviewed the great work she has being doing around capturing materials from listed buildings. This raises a few questions on the above which need a review. For historic buildings we often have more than two primary materials due to incremental additions and organic growth through time. When we discussed yesterday we said we'd start by uploading just buildings that have one or two materials to get something onto the map. Our option then might be something like
- More columns called tertiary material 
- Single extra column which allows a list of extra materials.

Theres also a question for Historic England here on what the materials column means and how it's generated, is the order in which they've given their materials the order of importance or just a list. Does glass ever mean window or a full glass wall?

polly64:
[...]

Issue #361, opened 2019-08-08
db: Add Construction/Materials category to database


dominijk:
**Fields for launch**
- Construction method > Single option autocomplete/drop down
- Primary material > Single option autocomplete/drop down
- Secondary material > Single option autocomplete/drop down (repeats primary)
- Primary roofing material > Single option autocomplete/drop down
- Secondary roofing material > Single option autocomplete/drop down (repeats primary)

**Steps for adding new fields to database;**

- [ ] Identify bulk data sources (optional)
- [x] Required - Define field values as static category page
- [ ] Required - Identify need for source and verification for fields, log to spreadsheet
- [ ] Required - Define input method for field (Drop down / Autocomplete / Free text) 
- [ ] Required - Add category down file
- [ ] Required - Add category up file
- [ ] Optional - Create nudge function (might be post launch) for linked fields as linked issue

On completion set up Github issue for activating fields on front end

MeldaS:
@polly64 I'm getting the code ready to launch the construction category. Just to confirm, you'd like to launch with just core_construction_material? 

If so, could you confirm / provide info for the following:

**Data field setting** -> Single option autocomplete/drop down (Basically a drop down menu with the list of materials. If users type directly into the box, suggested autocompleted text will appear). I think this is the ideal setting for Core Materials.

**List of materials:** (taken from issue #539)
- Wood #96613b
- Stone #ffffe3
- Brick #f5d96b
- Steel #beffe8
- Reinforced Concrete #fca89d
- Other metal #5c8970
- Other natural material #b5a859
- Other man-made material #c48a85

**Editable** -> should it be editable to the public? I would say yes. 

**Tool Tip** -> The little "i" button that provides info when you click on it. Could you help write the statement? Maybe something explaining that the core material is not necessarily the most prevalent one on the facade (if that's true). Or anything else that you think should be mentioned. 
"This is the core construction material of this building. xxx"


These are other data fields that we could also launch under the construction category. We could have them visible but not yet editable. Or just launch them and make the editable. If you'd like to launch any of the below, then we need to provide all the info above for each of the below. Here are my thoughts on each. 

- _Construction method_ > MS: I don't have any data to upload for this. Not sure if anyone else does.
- _Secondary material_ > MS: In the HE dataset on listed buildings, there are many secondary materials that we can list, but we'd need to think about if only one secondary material can be listed, or if we should allow users to select several from a list (my suggested option), or some sort of free text with validation (not sure how this would work)
[...]

PR #586, opened 2020-04-01
Feature/568 activate construction

MeldaS:
Created SQL scripts and modified relevant files according to [Adding new building attribute fields](../docs/adding-new-fields.md) file. 

**Relevant issues**: #568, #539 and #361 

**Added fields**: (taken from issue https://github.com/tomalrussell/colouring-london/issues/361#issuecomment-604657095)

- 'Core construction material'
- 'Secondary construction material/s'
- 'Main roof covering'

**Construction materials list & colours**: (taken from issue https://github.com/tomalrussell/colouring-london/issues/539#issuecomment-597298074)

- Wood #96613b
- Stone #ffffe3
- Brick #f5d96b
- Steel #beffe8
- Reinforced Concrete #fca89d
- Other metal #5c8970
- Other natural material #b5a859
- Other man-made material #c48a85

**Roof covering list**: (taken from issue https://github.com/tomalrussell/colouring-london/issues/361#issuecomment-604657095)

- Slate
- Clay tile
- Wood
- Asphalt
- Iron or Steel
- Concrete/Cement
- Other metal
- Other natural material
- Other manmade material

**Question**: I used SelectDataEntry for all three fields in the [Construction.tsx](../data-containers/construction.tsx) file. However, I'm not sure how to enable multiple selection only for the "Secondary Construction Material/s"


migrations/0xx.construction-materials.up.sql
```diff
@@ -0,0 +1,38 @@
+-- Launch of Constuction category
+-- Fields: Core construction material, Secondary construction materials, Roof covering materials
+
+-- Construction materials: Wood, Stone, Brick, Steel, Reinforced Concrete, Other Metal
+--   Other Natural Material, Other Man-Made Material
+CREATE TYPE construction_materials
```
mz8i, line 6: I'd change type name to `construction_material` (singular)

migrations/0xx.construction-materials.up.sql
```diff
@@ -0,0 +1,38 @@
...
+    'Wood',
+    'Asphalt',
+    'Iron or Steel',
+    'Other Metal',
+    'Other Natural Material',
+    'Other Man-Made Material');
+
+-- Core Construction Material
+ALTER TABLE buildings
+    ADD COLUMN IF NOT EXISTS core_materials construction_materials;
```
mz8i, line 30: if this is intended to hold just one value, let's rename column to `core_material` (singular)

migrations/0xx.construction-materials.up.sql
```diff
@@ -0,0 +1,38 @@
...
+
+-- Secondary Construction Materials
+ALTER TABLE buildings
+    ADD COLUMN IF NOT EXISTS secondary_materials construction_materials;
```
mz8i, line 34: Do we intend this to hold just one value? Secondary materials seems like something which could potentially hold multiple values (as opposed to core material) but then we need to change the column type to array of `construction_material`, which then necessitates several changes elsewhere in the code. Not sure what is the plan here.
If singular is ok, then I'd also change the column name to singular

migrations/0xx.construction-materials.up.sql
```diff
@@ -0,0 +1,38 @@
...
+CREATE TYPE roof_covering
+    AS ENUM ('Slate',
+    'Clay Tile',
+    'Wood',
+    'Asphalt',
+    'Iron or Steel',
+    'Other Metal',
+    'Other Natural Material',
+    'Other Man-Made Material');
+
```
mz8i, line 27: We've been discussing with @tomalrussell  that for classifications which aren't really a gold standard, but rather something we come up with - it might be better to use reference tables to define the list of acceptable values, rather than defining an `ENUM`. This is because enums are harder to change at a later date, and with those fluid types of classifications there might be a need for adding/removing options later on. This is as opposed to something like listing grade, where we most probably will never go beyond `I`, `II*` and `II` so enum is appropriate there.
However, to avoid additional work, I'd stick with `ENUM` here. Worst case, if we need to, we can swap the underlying mechanism later.

migrations/0xx.construction-materials.up.sql
MeldaS: Secondary materials should hold more than one value but I couldn't figure out how to create a drop down list than allows multiple selections. Any suggestions?

MeldaS: Secondary materials should hold more than one value but I couldn't figure out how to create a drop down list than allows multiple selections. Any suggestions?

mz8i: For the database side, the type of the column needs changing to an array type: `construction_material[]` 

The UI bit will be a `MultiDataEntry` component, but the latest version of that is on the land use branch so @tomalrussell needs to merge that one into master first. I thought we were going to just add the core material for construction, which is why we didn't go through that in the workshop. Currently the MultiDataEntry allows for a multi-valued list of free text fields, with an optional auto-fill dropdown which works by contacting the server the get the list of options based on typed text. There isn't currently an option for a pure dropdown option so that gets complicated. An idea for @tomalrussell: could change MultiDataEntry to accept any type of single data entry, for example through a render prop? So that it can be a list of dropdowns, list of text fields, list of text fields with autofill etc.

The amount of work is hard to estimate for the above - might be easiest to release just the single-valued core material and roof material, and then develop the secondary materials later?

migrations/0xx.construction-materials.up.sql
tomalrussell: Agreed - happy to start with the `ENUM`

migrations/0xx.construction-materials.up.sql
tomalrussell: Yes - this is a good note. I'll merge this as-is and see if it's straightforward to build on top with `MultiDataEntry` before releasing.

tomalrussell:
TODO after merge - come back to multiple secondary materials.

Issue #584, opened 2020-02-27
SIZE & SHAPE


dominijk:
Live  to do list

Then 
1. [ ] Reload UCL Energy height data as a lot missing
2. [ ] Reload UCL adjacency data as a lot missing
3. [ ] Create hint link to info page for height to indicate how tall buildings - PH to alter wix page colouring-cities/colouring-london#83
4. [ ] Create temporary hint links to Info page for roof shapes where images of  will be shown
5. [ ] Activate roofshape dropdown as below
6. [ ] Activate all other categories

1. Flat
2. Lean to / monoptich
3. Pitched gable
4. Pitched hipped 
5. Curved
6. Mansard
7. Gambrel
8. Complex pitched (saw tooth) 

Discussion

- Discuss how to add image  to dropdowns.  `DataEntry` type, radio button with images
- Possibly rethink type section 
- Think as to how 3D rule-based typology description could be added in time by creating link to FR's open procedural model site

Links colouring-cities/colouring-core#548 colouring-cities/colouring-core#556 , colouring-cities/colouring-core#434

polly64:
@dom FYI only
Think we need to chat more about the list and on this one will really need pics, Robert Hecht has been testing pics and crowdsourcing  at IOER Dresden

first stage revisions as follows
Pitched - gable
Pitched - hipped
Pitched - mansard with butterfly
Pitched - other (e.g saw tooth, gambrel)
Monopitch/lean-to  -(I have added hyphen here to make clearer 'lean-to')
Flat
Curved
Corrugated
Other

I have omitted gambrel as rare in uk and it can also be put under 'pitched other'

polly64:
copied across from 

Update legend as #434
 Add user tip to the legend to help users understand what legend means in storeys ie. a 20 metre building. could this be a small icon of buildings (house, small block flats, gherkin/shard)
 Check FAR is in db
 Activate FAR (adding min max constraints)
 Check plot area in db
 Activate plot area on front end (adding min max constraints)
 Remove configuration field
 Create new issue from discussion for roof shape, will be via new DataEntry type, image (of roof forms) with radio buttons #584
 Roof shape will sit under new toggle

polly64:
1. Address issue of missing matches for adjacency data 
2. Errors picked up in adjacency data
2. same with height data raised #556 - need to reload

PR #590, opened 2020-04-09
WIP for #361 intial construction db elements

tomalrussell:
- Materials list from #539
- Adding roof form for #584


tomalrussell:
TODO - revisit after #568 - may drop this file.

PR #665, opened 2021-02-26
Simple dynamics

mz8i:
A simple dynamics category, recording a list of records about past buildings on the site of the present building.
Changes include refactoring to data definition and validation code on the backend.

Things still missing:
- proper formatting of edit history
- better frontend form validation
- better UI styling
- better handling of changing order of records when editing


mz8i:
Replaced by #666

PR #670, opened 2021-04-26
Update core dependencies (2021/04)

mz8i:
Update core dependencies to latest version:
- razzle
- webpack
- typescript
- eslint
- react + react-dom
- leaflet
- react-leaflet
- more... (fixing npm audit warnings etc)

The update of react-leaflet also required changes to how the server-side leaflet problem is handled.
We switch from using `react-leaflet-universal` (which introduces it's own complications, and isn't available for react-leaflet v3) to using `@loadable/component` (a solution recommended by the React team). This setup will only load the `ColouringMap` component on the client-side, and so there's no need to wrap react-leaflet itself to prevent leaflet from being loaded on the server.
This also means that any custom react-leaflet (v3) components can now be used without any hacks, which potentially enables us to import a lot of useful leaflet libraries.


tomalrussell:
This all looks good.

lodash seem to be changing what they recommend - https://lodash.com/per-method-packages - as I read it, we should be doing:
```
import _isEmpty from 'lodash/isempty';
import _isEqual from 'lodash/isequal';
```

See #671 for another major bump, looks like we may be in a good position to update to node 14 (current LTS).

PR #671, opened 2021-04-27
Update pg-promise

tomalrussell:
Testing on node 14, pg-promise needs a moajor version bump.

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -2910,9 +2910,9 @@
       }
     },
     "assert-options": {
-      "version": "0.1.3",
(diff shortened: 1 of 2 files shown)


mz8i:
Definitely good idea, although would need to check if all the dependencies are easily installable - I think last time we tried, mapnik didn't have binaries available for that version. But it could be that after the dependency updates, this will no longer be the case. Would be keen to wrap up the current wave of updates with Node and Postgres!

mz8i:
For pg-promise, let's test that there are no breaking changes in categories that use some of the more advanced features of pg-promise, notably dynamics category

tomalrussell:
Closing this as superceded by #776

Issue #654, opened 2021-01-30
Add pure black background so have 3  options


polly64:
Can we add black background as well as existing dark so we have 3 options

Background options
1) Light - with plot outlines (best for editing)
2) Dark -with plot outlines
3) Black - no outlines (good for display)

tomalrussell:
Start in the code with https://github.com/colouring-london/colouring-london/blob/master/app/src/frontend/map/map.tsx#L144

PR #676, opened 2021-07-17
Night theme (Issue #654)

vincentschaffer:
Issue [654](https://github.com/colouring-london/colouring-london/issues/654)

**Overview**
We want to reintroduce the night theme since the Maps OS API has deprecated 'Night' [layer](https://osdatahub.os.uk/docs/wmts/technicalSpecification). We can just add a custom filter, as per their [docs](https://labs.os.uk/public/os-data-hub-examples/).

**Visuals**
Screen Shot 2021-07-17 at 6 01 10 PM

### DIFF
--- a/app/src/frontend/map/layers/city-base-map-layer.tsx
+++ b/app/src/frontend/map/layers/city-base-map-layer.tsx
@@ -9,25 +9,30 @@ const OS_API_KEY = 'UVWEspgInusDKKYANE5bmyddoEmCSD4r';
  * Base raster layer for the map.
  * @param theme map theme
  */
-export function CityBaseMapLayer({theme}: {theme: MapTheme}) {
+export function CityBaseMapLayer({ theme }: { theme: MapTheme }) {
 
     /**
      * Ordnance Survey maps - UK / London specific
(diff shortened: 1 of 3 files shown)


mz8i:
Thank you for the PR @vincentschaffer - the CSS filter is a very nice and simple idea!

I will test the style locally to see how the colour works with the buildings base layer (I assume you don't have access to the London building polygons so it's hard to test this independently).

In the meantime, a couple of very simple comments that would be great to address before merging:
- we haven't really been consistent with enforcing a certain style across the project, but a generally good idea is to follow the local style as much as possible - could you look through the changes and adjust this (I've noticed some snake_case instead of camelCase, double instead of single quotes etc)
- I can see quite a lot of whitespace edits in the diff, which I assume come from some auto-formatter you use on save. While I think it'll be good to re-format the whole project using a consistent style at some point, for this PR let's try to limit the changes to places where the logic is actually changed, so that it's easier to see the important parts of this change

Thanks again!

tomalrussell:
Thanks for this PR, @vincentschaffer and apologies it's taken so long to review. Closing in favour of #696 with a couple of edits on top of this.

Issue #681, opened 2021-08-07
land use not working


polly64:
@mz8i Land use feature not working -changed 'dwelling' to 'pub' for 52 lloyd baker street - option changed though save button didn't seemed to register but colour didn't change and the 'order' stayed residential

PR #684, opened 2021-08-12
Allow non-editable fields to be auto-derived

mz8i:
Some fields shouldn't be editable through the API
but can still be modified, because they are derived
from another field which is editable.
This change fixes a bug where a derived field
wouldn't be updated, because it was not on the
editable fields whitelist.


mz8i:
Closes #681.

PR #692, opened 2021-11-07
Document naming style

matkoniecz:
It appears to be not documented currently.

@mz8i Have I missed/misunderstood something?

Is there preference between `feature/document_style` and `feature/document-style`?


matkoniecz:
Note: I have no preferences here, but I wanted to document what is preferred so I can check it in future.

mz8i (approved):
That's a useful addition to the docs, thanks. This is a personal preference but I think the `feature/document-style` format looks cleaner

tomalrussell (approved):

PR #706, opened 2021-12-26
Display unverified residential differently

matkoniecz:
Sadly, there is still a remaining issue: verifying building is not causing map to be updated.

Note that `verifyBuildingAttributes` is actually invalidating cache (a bit too eagerly, but I have not investigated that).

@mz8i Do you have any idea what may be going on here?

screen07

### DIFF
--- a/app/map_styles/polygon.xml
+++ b/app/map_styles/polygon.xml
@@ -530,7 +530,11 @@
             <PolygonSymbolizer fill="#cccccc" />
         </Rule>
         <Rule>
-            <Filter>[current_landuse_order] = "Residential"</Filter>
+            <Filter>[current_landuse_order] = "Residential" and [verified_residential] = 0</Filter>
+            <PolygonSymbolizer fill="#6c6f8e" />
+        </Rule>
+        <Rule>
+            <Filter>[current_landuse_order] = "Residential" and [verified_residential] = 1</Filter>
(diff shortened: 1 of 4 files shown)


matkoniecz:
Deployed for testing on https://cl-staging.uksouth.cloudapp.azure.com/

matkoniecz:
I missed some extra trigger needed, beyond triggering cache expiry, to redraw map.

Note how verifying fails to change colour (despite logging tile expiry) and how changing building type trigger redraw.

Peek 2021-12-26 04-00

matkoniecz:
@polly64 What you think about colour used and legend description?

polly64:
hi @matkoniecz think the blues need to be better differentiated maybe the verified one significantly lighter?
Sorry didn't quite understand your note above re the problem with changing colour when verified- does your comment mean you're still working on it or you think there is a problem in making it work t all? hope you had great christmas

polly64:
@matkoniecz i realised after we spoke that in the next stage we still also want to find a way of colour visualising which entries (in any of the 12 categories) have been verified and which not, so if you have any thoughts let's chat anytime

polly64:
@matkoniecz don't worry it's fine we can look at this once other stuff complete- it just about looking at ways to show coloured buildings have been verified but i think we need to think and test this over time

matkoniecz:
being replaced by a better PR, will be send soon

PR #710, opened 2022-01-14
Enable additional zoom levels on the map view

matkoniecz:
Note: in the previous attempt there were some issues on retina devices. It should be tested before merging, even if it works on some other devices.


matkoniecz:
borrower retina device: and yes, it is broken there

next step: upgrading dependencies, retest

PR #713, opened 2022-01-14
Link provision script

matkoniecz:
Not entirely sure where it should be linked (maybe this entire section should be moved to CONTRIBUTING.md) but more prominent linking seems to be a good idea to me.


edwardchalstrey1:
Looks good, I think it's nice for things to be findable on the README repo homepage

PR #715, opened 2022-01-17
add edwardchalstrey1 to contributors

edwardchalstrey1:
Testing I can commit, PR, merge, etc

PR #717, opened 2022-01-17
Revert "add edwardchalstrey1 to contributors"

edwardchalstrey1:
Reverts colouring-cities/colouring-london#715


matkoniecz:
I guess that it would be nice to have some docs how this works in https://github.com/colouring-cities/colouring-london#contributors (I knew that bot was involved but assumed that it updates rendering based on JSON specification)

PR #735, opened 2022-01-19
unbreak python3.8

matkoniecz:
https://github.com/psycopg/psycopg2/releases/tag/2_8_4

step toward working 20.04 Vagrant

Note: #705 introduced mismatch between `provision/vm_provision.sh`  and https://github.com/colouring-cities/colouring-london/blob/master/Vagrantfile#L15

This would bring it closer to working 20.04 Vagrant but that would break 18.04 Vagrant even more ("PostgreSQL at least 11.4"). Maybe it would be better to have done in one step (but #705 happened already).


matkoniecz:
will be replaced by more complete one

PR #769, opened 2022-01-31
hide likes

matkoniecz:
implements second part of #734 (hiding them completely was discussed on the meeting)

Issue #734, opened 2022-01-19
Facilitate building attribute data upload for Community


edwardchalstrey1:
@matkoniecz @edwardchalstrey1  CLHEAG PRIORITY

The main purpose of the category is to:
- identify the type of buildings viewed as being of value to local communities and provide  relevant geospatial data for analysis 
- raise awareness of building threatened with demolition
- track/quantify the scale of loss/additions of community assets over time
- quantify and geolocate all buildings in non-profit ownership in the city 
- provide a tool to allow communities to change demolition of important local buildings

Actions required

- [ ] Allow to save edits, doesn't currently work https://github.com/colouring-cities/colouring-london/issues/734#issuecomment-1025690548
- [x] Remove  the 'x people like this building' x 2 completely 
- [x] Change the question 'Do you think this building should be recorded as one of special local significance?' to read 'Do you think this building should be recorded as a local heritage asset? (add x ticks)
- [ ]  Add new subcategory  'Is this building already registered as being of local or national significance?'
Yes/No 
- [ ] Add link to Planning category add info on Planning category (do this one later)
- [x] Change 'Can you help add information about public ownership of the building' to 'Can you help add information on community use of buildings?'
- [x] Change the title to  'Is building in some sort of "community" ownership'" Yes/No (covered by #774)
- [x] Change above dropdown wording from 'state-owned' to 'government owned'
- [x] Add 'Are activities open to the community currently taking place in the building?' Yes/No  (covered by #774)
- [x] Add 'Has the building always been used for community activities?' Yes/No  (covered by #774)
- [x] Add 'Has the building ever been used in the past for community activities, but isn't now? Yes/No  (covered by #774)

#773 (comment)

- [ ] #773
- [ ] Add copy tool 

For history see also closed issue #602

polly64:
The main purpose of the category is to:
a)  identify the type of buildings viewed as being of value to local communities and provide  relevant geospatial data for analysis 
b)  raise awareness of building threatened with demolition
c)  track/quantify the scale of loss/additions of community assets over time
d) quantify and geolocate all buildings in non-profit ownership in the city 






see also closed issues

matkoniecz:
@polly64 

> Change the 'x people like this building' to not include 0 likes.

Can be done but not entirely sure is it improving consistency of interface.

>  Change the 'x people like this building to '2 ticks' not '2 likes')

[...]

PR #771, opened 2022-02-01
Feature/reword local asset

matkoniecz:
> Change the question 'Do you think this building should be recorded as one of special local significance?' to read 'Do you think this building should be recorded as a local heritage asset? (add x ticks)

from #734

draft, as #769 needs to go first

screen02


matkoniecz:
@polly64 Is screenshot above matching what was expected?

polly64:
@matkoniecz  yes thanks

PR #772, opened 2022-02-01
Better describe Community section

matkoniecz:
better describe section about community use

Finally something as easy as expected.

Here just to give @polly64 a chance to check

from #734 covers

>  Change 'Can you help add information about public ownership of the building' to 'Can you help add information on community use of buildings?'

>  Change 'Has the building ever been used for community activities' to 'Has this ever been used for community activities in the past'

screen


polly64:
@matkoniecz i think this is ok for the moment

PR #780, opened 2022-02-14
WIP: Dockerise Colouring London

edwardchalstrey1:

PR #781, opened 2022-02-14
[WIP] Fix vagrant virtual machine deployment

edwardchalstrey1:
Note: duplicates some of the changes in #762

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -12,7 +12,7 @@ Vagrant.configure("2") do |config|
 
   # Every Vagrant development environment requires a box. You can search for
   # boxes at https://vagrantcloud.com/search.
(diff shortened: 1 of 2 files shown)


matkoniecz:
Are you stuck on something specific here? (I was also stuck on my Vagrant attempt, specifically on Mapnik building - again - but I may try looking at it again)

edwardchalstrey1:
> Are you stuck on something specific here? (I was also stuck on my Vagrant attempt, specifically on Mapnik building - again - but I may try looking at it again)

See my comment on #780 - I will continue work on docker first, but this may also be a nice thing to have at some point - since the docker version does not have problems with Node, if we get that working first, perhaps we can transpose some of the same changes to this context in case people wish to use Vagrant

PR #783, opened 2022-02-16
Replace State-owned by Government-owned

matkoniecz:
part of #734


mz8i (commented):
The code looks ok, though it's a lot to just change an enum value - i think the general conclusion is that for list-of-options type fields, we should strive for using string types rather than enums, and doing the validation in the application layer. This has actually been the general direction, apart from this introduction of community fields where we've reused SQL that was written a while back.

I'm not sure if I understand the difference between state-owned and government-owned, though. Are the Houses of Parliament government-owned? Or state-owned?

Issue #727, opened 2022-01-19
Facilitate building attribute data upload for Construction


edwardchalstrey1:
### Easy
- [ ] Add 'main secondary construction materials' list and allow them to tick however many are relevant. Use primary construction list (partial overlap with #568)
- [x] Change name from 'secondary construction materials to 'main secondary construction material/s'
- [x] Change sub category name from 'construction system' to 'construction system type'
- [ ] Add 'main construction system' list dropdown as follows:
- solid wall
- frame construction
- structural prefabricated panels 
- shell or membrane structures
- tensile structures 
- other (if other chosen provide free text box saying please specify
- [ ] Add 'source' box but allow multiple links to be added
-  [x] Swap the colour allocated for 'wood' for the colour allocated for 'other natural material' as a) the wood colour is too dark on black background, b) unsatisfying to colour in  (steel and RC colours very good/satisfying to use and brick ok, and  c) as so few wood buildings
Current colour key
name	rgb	hex
Timber (Light Brown) | #96613b
Stone (Cream) | #b5a859
Brick (Bright yellow) | #fca89d
Steel (Pink) | #5c8970
Glass (cream) | #ffffe3
Concrete (turquoise) | #beffe8
Other metal | #4d0000
Other natural material | #96613b
Other man-made material | `#c48a85'
- [ ] Add copy tool for each & a copy all option
- [ ] Add verification button for each

### Harder

- [ ] Download building age data, classify all pre 1900 buildings as main construction type = brick, and main construction system = solid wall, and upload. (several 100,000 might be worth asking Ed this)
- [ ] Add multiple visualisation options - for each of the 4 data types collected - as Maciek has done for Community category (can be done after feb)

### Other notes for PH

Explore collection of cladding and cladding date info 
Historical GiThub refs: 
https://github.com/colouring-cities/colouring-london/issues/605 #539

matkoniecz:
@polly64 What needs to be changed here?

polly64:
@matkoniecz just adding things now

matkoniecz:
@polly64 

Are

> Secondary Construction Material/s
> Construction system

sections working currently in any way or were ever working? Not sure is it just placeholder or something got broken (I can check in repo history but if you remember it well it would be helpful)

polly64:
@matkoniecz no secondary materials were never working we just had placeholder

matkoniecz:
It is partially overlapping with #361

matkoniecz:
[...]

PR #785, opened 2022-02-18
stop showing ids in multiselect dropdown

matkoniecz:
it is not really needed for users, not critical for debugging, showing this ids is not helpful and rather confusing
especially as you cannot search for entries using them
it gets worse for #727 (secondary material multiselect dropdown) as materials have only names, not ids

screen

before this change WIP for #727 had following display (landuse mentioned here in prefill is obviously also going to be fixed):

screen01

### DIFF
--- a/app/src/frontend/building/data-components/autofill-dropdown/autofill-dropdown.tsx
+++ b/app/src/frontend/building/data-components/autofill-dropdown/autofill-dropdown.tsx
@@ -60,7 +60,6 @@ export const AutofillDropdown: React.FC<AutofillDropdownProps> =  props => {
             {
                 options.map(option =>
                     <div
-                        key={option.id}
                         onMouseDown={e => /* prevent input blur */ e.preventDefault()}
                         onClick={e => {
                             props.onSelect(option.value);
@@ -68,7 +67,7 @@ export const AutofillDropdown: React.FC<AutofillDropdownProps> =  props => {
                             props.onClose();
                         }}
                     >
-                        {option.value} ({option.id})
+                        {option.value}
                     </div>)
             }
         </div>
--- a/app/src/api/services/autofill.ts
+++ b/app/src/api/services/autofill.ts
@@ -1,7 +1,6 @@
 import db from '../../db';
 
 interface AutofillOption {
-    id: string;
     value: string;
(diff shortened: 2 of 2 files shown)


mz8i (changes requested):
Whether the land use IDs are useful or not is debatable - one could argue that because we're using a standardised classification, it is good to display the official code of each land use, for users to be able to look it up easily.

I agree that the inability to search the list by typing the code is a problem.

Anyway, we'll need some sort of ID because of React's requirement in the case of a list of elements (see review comment in code). The two options are:
- leave the ID as-is in the sql query and the react list. For #727, return the material type as both ID and label (plus, implement showId prop as described below)
- remove the ID from the query and use the label as the react key (only do that if we think there's genuinely no use in showing the land use ID)

Another potential change is to keep the ID but add a boolean "showId" prop to the autocomplete component and add the ID text in parentheses based on the value of that prop.

Re the comment on #727 having land use text as placeholder - is the placeholder text currently hard-coded in the autocomplete component? If so, we need to expose it as a prop. But that's a separate issue

app/src/frontend/building/data-components/autofill-dropdown/autofill-dropdown.tsx
```diff
@@ -60,15 +60,14 @@ export const AutofillDropdown: React.FC<AutofillDropdownProps> =  props => {
...
                     <div
-                        key={option.id}
                         onMouseDown={e => /* prevent input blur */ e.preventDefault()}
                         onClick={e => {
                             props.onSelect(option.value);
                             // close dropdown manually rather than through input blur
                             props.onClose();
                         }}
                     >
-                        {option.value} ({option.id})
```
matkoniecz, line 71: @mz8i So, another option would be to query as so far - but simply to stop showing it to user, right?

tomalrussell (changes requested):
As with @mz8i comments, we need some `key` to keep React happy.

I think here there is some small benefit to interested users to see the ID - I might even argue for showing the ID in the inactive data entries / view display, though that would be a bigger change and not in scope here.

I think it's a good idea to add a `showId` property if you want to use this in multiple contexts, some with id, others without.

PR #790, opened 2022-02-23
Add support for secondary construction materials

matkoniecz:
part of #727

### DIFF
--- a/app/src/api/services/domainLogic/secondaryMaterial.ts
+++ b/app/src/api/services/domainLogic/secondaryMaterial.ts
@@ -1,0 +1,29 @@
+import _ from 'lodash';
+
+import * as materialDataAccess from '../../dataAccess/material';
+import { ArgumentError } from '../../errors/general';
+
+export interface materialGroup {
+    materials: string[];
+}
+
+export async function updateSecondaryMaterials(update: materialGroup): Promise<materialGroup> {
+    const groupMaterials = update.materials;
+
+    for(let group of groupMaterials) {
+        const isGroupValid = await materialDataAccess.isMaterialAllowed(group);
+        if(!isGroupValid) {
+            throw new ArgumentError(`'${group}' is not a valid Land Use Group`, 'secondaryMaterialsUpdate');
+        }
+    }
+
+    if(hasDuplicates(groupMaterials)) {
+        throw new ArgumentError('Cannot supply duplicate materials', 'secondaryMaterialsUpdate');
+    }
+
+    return update;
+}
+
+function hasDuplicates(array: any[]) {
+    return (new Set(array).size !== array.length);
+}
--- a/app/src/frontend/config/data-fields-config.ts
+++ b/app/src/frontend/config/data-fields-config.ts
@@ -346,7 +346,7 @@ export const dataFields = { /* eslint-disable @typescript-eslint/camelcase */
         category: Category.Construction,
         title: "Main Secondary Construction Material/s",
         tooltip:"Other construction materials",
-        example: "",
+        example: ["", ""],
     },
 
     construction_roof_covering: {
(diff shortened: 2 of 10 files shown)


matkoniecz:
Note that primary materials are not appearing to be validated and just rely on https://github.com/colouring-cities/colouring-london/blob/master/app/src/frontend/building/data-containers/construction.tsx#L10 being in synch with database (I think?).

See also https://github.com/colouring-cities/colouring-london/blob/master/app/src/frontend/building/data-containers/community.tsx#L127 repeating `public_ownership_type` enum.

I am not entirely sure is this database-based verification is needed. Maybe it should be limited to dropdown selection and block freeform entries? Maybe it is superior to relying on this manually maintained copy of valid value list?

edwardchalstrey1:
I have tried to take a look at this but it creates an error in the app for me. I did the following:

1. Check out this branch
2. Run `psql -f migrations/027.secondary_materials.up.sql` (I have the db env var set to "colouringlondondb") - this ran fine
3. `npm start`:

```bash
ed@ed-VirtualBox:~/colouring-london/app$ npm start

> colouring-london-app@1.0.0 start
> razzle start

 WAIT  Compiling...


✔ Client


✔ Server
  Compiled successfully in 13.91s

Issues checking in progress...
ERROR in src/frontend/route.tsx:36:25
TS2339: Property 'from' does not exist on type '{}'.
    34 |         render={props => {
    35 |             if(isAuthenticated) {
  > 36 |                 const { from } = props.location.state ?? { from: '/my-account.html'};
       |                         ^^^^
    37 |                 return <Redirect to={{pathname: from }} />;
    38 |             } else {
    39 |                 if(Component) {

sswp> Handling Hot Module Reloading
✅  Server-side HMR Enabled!
🚀  started
```

edwardchalstrey1:
@matkoniecz I realise that the error I posted above might be some basic typescript error that I don't understand without looking into it more `Property 'from' does not exist on type '{}'`, I get the same error on #785 & #791 so perhaps is there some easy step I can take here? I will look into it further if it's not immediately obvious to you 😄

tomalrussell:
> @matkoniecz I realise that the error I posted above might be some basic typescript error that I don't understand without looking into it more `Property 'from' does not exist on type '{}'`, I get the same error on #785 & #791 so perhaps is there some easy step I can take here? I will look into it further if it's not immediately obvious to you smile

https://github.com/colouring-cities/colouring-london/commit/4e40b31bacfd59b65936f3f792554f2b925a7152 should do it

app/src/api/services/domainLogic/secondaryMaterial.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    for(let group of groupMaterials) {
+        const isGroupValid = await materialDataAccess.isMaterialAllowed(group);
+        if(!isGroupValid) {
+            throw new ArgumentError(`'${group}' is not a valid Land Use Group`, 'secondaryMaterialsUpdate');
+        }
+    }
+
+    if(hasDuplicates(groupMaterials)) {
+        throw new ArgumentError('Cannot supply duplicate materials', 'secondaryMaterialsUpdate');
```
edwardchalstrey1, line 21: Looks good to me, I have just one comment - perhaps similar to this error thrown here, there can be one for when the Main Secondary Construction Material/s selected is the same as the Core Material (presumably we don't want that)

edwardchalstrey1 (approved):

app/src/frontend/config/data-fields-config.ts
```diff
@@ -359,7 +359,7 @@ export const dataFields = { /* eslint-disable @typescript-eslint/camelcase */
     sust_breeam_rating: {
         category: Category.Sustainability,
         title: "BREEAM Rating",
-        tooltip: "(Building Research Establishment Environmental Assessment Method) May not be present for many buildings",
+        tooltip: "(Building Research Estacblishment Environmental Assessment Method) May not be present for many buildings",
```
tomalrussell, line 362:
```suggestion
        tooltip: "(Building Research Establishment Environmental Assessment Method) May not be present for many buildings",
```

tomalrussell (approved):
Just suggested a fix for typo which crept in, otherwise good to go.

I'd be open to a separate discussion of data typing and storage. I think this approach works okay, but we are incrementally adding complex data types to our one big table. Could consider moving towards more normalised tables, (or another direction, e.g. towards JSON blobs) - but would suggest doing that with a look across the data model as a whole.

Issue #722, opened 2022-01-19
Update Colouring London 2016 OSMM polygons and facilitate automatic updating


edwardchalstrey1:
### Tasks

- [x] Request to download the data from [here](https://www.ordnancesurvey.co.uk/business-government/licensing-agreements/data-exploration)
    - [x] Download data once request received
- [x] Document existing DB setup with latest OS data #798 
- [x] Create script for adding the new data to db (see below comment) & document the script that does this update, test in a dev environment. Set up a db from dump file and then update it with new script.
- ~~Come up with a way in the app code to view older versions (previous OS release years or historical map years), test in dev~~ #751 <- lets do this separately later
- [x] #804
- [x] Roll out the update to CL servers

## Linked

- #817
- #818

edwardchalstrey1:
### Facilitate automatic updating

Script should:

- [ ] Add polygons from new OS data to geometries table that have **new** TOIDs i.e. are not already in the db & create new building record to match, with best guess build date of today
- [ ] Check if any of the features (e.g. lat/lon) in the update (the new geometries) match one of our pre-existing TOIDs (pre-update geometries) but have a different geometry/shape - in which case mark the old one as valid to today and add the new one as valid from today, keep both linked to the existing building record (so don't have a new building record when this is the case)
- [ ] Check if any of our pre-existing TOIDs are no longer present in the update - in which case mark as valid to today, and assume the building no longer exists, with best guess demolition date of today

Issue #800, opened 2022-03-14
Document how polygons get from the database to the map


tomalrussell:
How do the polygons go from being stored in the database to being visible on the map once the app is running?

The entry point for the server application is:
- [`/app/src/index.ts`](https://github.com/colouring-cities/colouring-london/blob/8c56d206388825ffe37c2fba71fe7be78b9eed10/app/src/index.ts)

This loads the "server" modules, which defines an ExpressJS application:
- [`/app/src/server.tsx`](https://github.com/colouring-cities/colouring-london/blob/8c56d206388825ffe37c2fba71fe7be78b9eed10/app/src/server.tsx)

This loads routes to handle API requests, in particular the tileserver, which parses URL params here:
- [`/app/src/tiles/tileserver.ts`](https://github.com/colouring-cities/colouring-london/blob/8c56d206388825ffe37c2fba71fe7be78b9eed10/app/src/tiles/tileserver.ts)

And calls to render the tile here:
- [`/app/src/tiles/rendererDefinition.ts#L74`](https://github.com/colouring-cities/colouring-london/blob/8c56d206388825ffe37c2fba71fe7be78b9eed10/app/src/tiles/rendererDefinition.ts#L74)

With a bit of indirection - checking cache, potentially stitching together a tile image from higher-zoom rendered images - we end up at a call or calls to create a tile image from the database. This uses the Mapnik library to connect to Postgres and fetch building geometries and data (using the injected getTableDefinitionFn and getVariablesFn), with reference to the style definition (loaded from XML) to render and return a tile image:
- [`app/src/tiles/renderers/renderDataSourceTile.ts#L32`](https://github.com/colouring-cities/colouring-london/blob/8c56d206388825ffe37c2fba71fe7be78b9eed10/app/src/tiles/renderers/renderDataSourceTile.ts#L32)

The style rules are defined in XML. These mostly just map from data values to colours, one Style declaration per tileset.
- [`/app/map_styles/polygon.xml`](https://github.com/colouring-cities/colouring-london/blob/8c56d206388825ffe37c2fba71fe7be78b9eed10/app/map_styles/polygon.xml)

The data sources and queries are defined here:
- [`/app/src/tiles/dataDefinition.ts`](https://github.com/colouring-cities/colouring-london/blob/8c56d206388825ffe37c2fba71fe7be78b9eed10/app/src/tiles/dataDefinition.ts)

For example, here's a single tile from the "date_year" tileset, with buildings coloured by year built:
[...]

PR #808, opened 2022-04-14
Add update scripts for Ordnance Survey data

### LINKED ISSUE #804: Set map boundary in intelligent way
Update `etl/drop_outside_limit.sh`, test it and add to documentation. Perhaps add an argument that specifies whether the boundary is London or Britain.

edwardchalstrey1:
Does most of #722 
Closes #804


app/src/tiles/dataDefinition.ts
```diff
@@ -7,12 +7,16 @@ const LAYER_QUERIES = {
         SELECT
             geometry_id
         FROM
-            buildings`,
+            buildings
+        WHERE
+            latest_demolish_date IS NULL`,
```
edwardchalstrey1, line 12: Hi @tomalrussell - I'm at the point where I have scripts that I can use to update the postgres db which I'm ready to try out on the staging server - one thing I still haven't correctly implemented is a change that will prevent buildings from the buildings table showing up when their `latest_demolish_date` is anything other than `NULL` (if you remember our discussion, that's the simple check we're going with for the time being). I've made an attempt here to update the ts code in the right place after looking at your explanation on #800 - this doesn't seem to be sufficient to prevent a building in the buildings table showing up in the app when is has a `latest_demolish_date` - any thoughts what else I need to do?

edwardchalstrey1: Thanks Tom, that's really helpful, I think this almost works correctly now - you were right it was doing something anyway, but now should be more general with the change I've made. I'm still getting the geometry id from "demolished" buildings showing up when I click on where they're located though. I'll investigate further, but this might not matter, so long as the polygon is non-editable in the website.

tomalrussell: That goes through the API `/api/buildings/locate?lng=...&lat=...`

from the frontend, the request is made in response to a map click:
- https://github.com/colouring-cities/colouring-london/blob/e12b544503a5e5ba302387a8c2bddee1a5be514d/app/src/frontend/map/map.tsx#L60

from the backend, here's the stack drilling down to where that is eventually handled in a database query: 
- entry point https://github.com/colouring-cities/colouring-london/blob/master/app/src/api/api.ts#L20
- server mounts API https://github.com/colouring-cities/colouring-london/blob/master/app/src/server.tsx#L47
- buildings routes loaded at https://github.com/colouring-cities/colouring-london/blob/master/app/src/api/api.ts#L20
- route defined at https://github.com/colouring-cities/colouring-london/blob/master/app/src/api/routes/buildingsRouter.ts#L12
- controller method at https://github.com/colouring-cities/colouring-london/blob/master/app/src/api/controllers/buildingController.ts#L16
- query at https://github.com/colouring-cities/colouring-london/blob/a4771eaac0717febee4e1e57b585ac6aca84c9e5/app/src/api/services/building/query.ts#L3

edwardchalstrey1: Brilliant thanks, editing query.ts did the trick

PR #825, opened 2022-05-20
Fix Current Use verification refresh issue

### LINKED ISSUE #831: Trigger refresh of app when verify button clicked for Current Use is residential
Trigger refresh of app when verify button clicked for Current Use is residential (like what happens when current use category group changed) or when any verify button is clicked if easier

edwardchalstrey1:
[WIP] @edwardchalstrey1 will come back to this

Closes #831

### DIFF
--- a/app/src/frontend/building/data-container.tsx
+++ b/app/src/frontend/building/data-container.tsx
@@ -247,9 +247,15 @@ const withCopyEdit: (wc: React.ComponentType<CategoryViewProps>) => DataContaine
                 this.setState({error: err});
             }
             
-            const edits = {
-                [slug]: value
-            };
+            if (slug == 'current_landuse_group'){
+              const edits = {
+                  ['current_landuse_verified']: TRUE
+              };
+            } else {
+              const edits = {
+                  [slug]: value
+              };
+            }
 
             this.doSubmit(edits);
             console.log(slug + " verify button clicked")
--- a/app/src/frontend/building/data-container.tsx
+++ b/app/src/frontend/building/data-container.tsx
@@ -248,7 +248,7 @@ const withCopyEdit: (wc: React.ComponentType<CategoryViewProps>) => DataContaine
             }
             
             const edits = {
-                [slug]: null
+                [slug]: value
             };
 
             this.doSubmit(edits);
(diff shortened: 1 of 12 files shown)


app/src/frontend/building/data-container.tsx
```diff
@@ -246,6 +246,13 @@ const withCopyEdit: (wc: React.ComponentType<CategoryViewProps>) => DataContaine
             } catch(err) {
                 this.setState({error: err});
             }
+
+            const edits = {
+                [slug]: value
+            };
+
+            this.doSubmit(edits);
```
edwardchalstrey1, line 254: @tomalrussell @matkoniecz I wonder if either of you could help me here - I'm trying to get it so that the page refreshes whenever a verify button is clicked, as is currently the case when an edit is saved. My slightly hacky attempt here has been to call the `this.doSubmit` within verification handling similar to edit handling.

Specifically the reason for this update is that I previously updated the Current Land use section with an extra colour for "Residential (verified)" to differentiate from buildings presumed Residential by default. The colour already changes when you click Verify under "Current Land Use (Group)", which is what we want, however it takes a refresh (or zooming in/out) to get the change to be visible. So having the page refresh on clicking Verify should solve this - I think it would be fine to have it refresh whenever *any* Verify button is clicked, but you may disagree?

tomalrussell: The mechanism to refresh the map tiles (without needing an actual page refresh) uses the `revisionId`, which is updated when building data is changed (and an entry is made in the edit history) but not when metadata is changed (and an entry is made in the user verification table).

You can trace how this is used through https://github.com/colouring-cities/colouring-london/search?q=revisionId 

In particular it gets added to the tilelayer URL [here](https://github.com/colouring-cities/colouring-london/blob/7343c71d169b09e82482c4f6bac06f7fb03edd2e/app/src/frontend/map/layers/building-data-layer.tsx#L10-L11) in the GET parameter string - this makes no difference to the backend, but it does force the browser to refetch map tile images from the server.

How do you think it would be best to incorporate verification (any possibly other metadata?) in this mechanism? Should we add a `verificationId` alongside `revisionId`? Or include the verification count in the cache-busting query parameters?

edwardchalstrey1: I think the simpler solution, since what we're trying to solve here is not common to verification in general, but specific to verifying the Current Land Use (Group), is to record this specific kind of verification in the edit history and so update the `revisionId` - which seems logical as the result is an "Edit" - a colour change. I've started looking into this on #833 - but we can chat about this this afternoon.

PR #834, opened 2022-06-06
Activate team section

### LINKED ISSUE #730: Facilitate building attribute data upload for Team
@matkoniecz @edwardchalstrey1 CLHEAG committee priority to activate

- [x] Add main Note to contributor in a blue box (using same format as for 'Current use') saying:
'Can you help us capture information on who built the current building?'

- [x] 1a. Add text 'Do you wish to add information on:
'The main building?' OR 'A major extension?'

This answer leads you to the same questions for each option:
- [x] 2a. Add question 'When was the work carried out?' 
(If this relates to the main building and not an extension, and the age already been provided under 'Age ', it should automatically prepopulate - as code for age in Dynamics, but also make sure if someone adds age for Team that it prepopulates the other way too i.e. in Age)
- [x] 2b. Add verify- to also verify at same time for same entry in  'Age'
- [ ] ~~2c. Add Source link but no verification for source (only allow urls to be entered)~~

- [x] 3a. Add question: 'What type of developer built the building?'  
(This is the dropdown for ownership as under 'Community' which i have just noticed  hasn't been activated- i thought it had could you check?)
a) State
b) Charity
c) Community/Cooperative
d) Other non-profit body
f)  Private (jndividual)
g) Private (company/estate)
h) Other

- [x] 4.a Add 'Who was/were the main designer/s?' (Free text. First name,  space, then Last name- include constraints to maximise security i.e. max number of characters) Use plus button as in land use to allow addition of up to five designers
- [x] 4b. Add verify
- [x] 4c. Add Source link but no verification for source
- [x] 4d. How best describes the lead designer?
Landowner
Speculative builder
Government architecture department
Architectural firm
Engineering firm
Other

- [x]  5.a Add 'Did the team win any awards for this building?' 

- [x] 5b. If yes 'Add source link' and plus button so you can add up to 5

edwardchalstrey1:
Closes #730 

Hi @polly64 I've updated the Team section as per your specs, however you'll definitely want to check how this looks on https://cl-staging.uksouth.cloudapp.azure.com/ before we roll it out.

In particular, I've made a few judgement calls:

1. Kept the "Year built" field (which is the same as the Age category) separate from "Year extension built", which only shows up if you click a "Is there an extension?" Yes/No
2. Added Verify buttons to everything to be consistent with the other categories

One thing that this doesn't cover is a situation where there is >1 extensions built in different years - is this something we want to include? Let's chat about this when you have time.

@tomalrussell @matkoniecz if you have any comments on the code let me know, thx - in particular the way I've handled the React components for the extension logic - see line 53 of `team.tsx`

### DIFF
--- a/app/map_styles/polygon.xml
+++ b/app/map_styles/polygon.xml
@@ -85,19 +85,19 @@
             <PolygonSymbolizer fill="#994d00" />
         </Rule>
         <Rule>
-            <Filter>[team_info_count] &gt;= 6 and [location_info_count] &lt; 8</Filter>
+            <Filter>[team_info_count] &gt;= 6 and [team_info_count] &lt; 8</Filter>
             <PolygonSymbolizer fill="#e67300" />
         </Rule>
         <Rule>
-            <Filter>[team_info_count] &gt;= 4 and [location_info_count] &lt; 6</Filter>
+            <Filter>[team_info_count] &gt;= 4 and [team_info_count] &lt; 6</Filter>
             <PolygonSymbolizer fill="#ff9933" />
         </Rule>
         <Rule>
-            <Filter>[team_info_count] &gt;= 2 and [location_info_count] &lt; 4</Filter>
+            <Filter>[team_info_count] &gt;= 2 and [team_info_count] &lt; 4</Filter>
             <PolygonSymbolizer fill="#ffbf80" />
         </Rule>
         <Rule>
-            <Filter>[team_info_count] &gt; 0 and [location_info_count] &lt; 2</Filter>
+            <Filter>[team_info_count] &gt; 0 and [team_info_count] &lt; 2</Filter>
             <PolygonSymbolizer fill="#ffe6cc" />
         </Rule>
 
--- a/app/map_styles/polygon.xml
+++ b/app/map_styles/polygon.xml
@@ -79,6 +79,39 @@
             <LineSymbolizer stroke="#888" stroke-width="3.0"/>
         </Rule>
     </Style>
+    <Style name="team">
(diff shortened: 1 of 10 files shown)


polly64:
@edwardchalstrey1  thanks made new account.
could you add
- [ ] 'religious body' to the landowner dropdown
- [ ]'architect/architectural firm' to the designer dropdown istead of  just 'architectural firm'

also 

- [ ] the building needs to colour the more data categories are filled - i.e. exactly same format as for 'Location' data- can you use green or orange colour scale instead of the blue, could u add copying that code?
- [ ] realise we need to be able to capture data on whole building and on all extensions at same time or separately . i.e. rather than having  an either/or - meaning you have to state the whole process again for each extension - any thoughts re quick fix for this?
- [ ] the verify button wouldn't work til i had saved other stuff , could u check?

polly64:
Hi @edwardchalstrey1 looking great. I've just been testing and this is list to come out of it 

GENERAL

- [ ] Stop digits moving so you actually have to type in number as v easy to accidentally alter and we don't really need them

LOCATION
- [ ] Activate all the boxes that are restricted but use the same protection as for name i.e. if address or building name maximum number of characters and need for gap between road name and word road/street/avenue etc

FOR TEAM

- [ ] Remove 'NO' option on awards 
- [ ] Change to text to read 'Who was/were the developer/s'
- [ ] Change developer dropdown to read ' Commercial company' (rather than Private company)
- [ ] Change Engineer to ' Engineer/Engineering firm'
- [ ] 

FOR PLANNING


FOR COMMUNITY

PR #840, opened 2022-06-20
Enable copy tool for all fields

edwardchalstrey1:


edwardchalstrey1:
@tomalrussell @matkoniecz I plan to look at this further later this week, but let me know if you see a quick solution to 2 in the top comment - I'm trying to make the checkboxes filled in by default, my current implementation with the `initCopyProps` doesn't break anything but doesn't work

tomalrussell:
In `/app/src/frontend/building/data-container.tsx` you could set `keys_to_copy` to default non-empty 
 [`getDerivedStateFromProps`](https://github.com/colouring-cities/colouring-london/blob/3140f575c310334e6bb83be780234c5356eb6eb6/app/src/frontend/building/data-container.tsx#L80-L95)

Things to think about: 
- default not to copy empty fields (would be surprising to copy empty values, effectively deleting data)
- include only fields visible within the current category - could use [`dataFields`](https://github.com/colouring-cities/colouring-london/blob/3140f575c310334e6bb83be780234c5356eb6eb6/app/src/frontend/config/data-fields-config.ts#L112) and `this.props.cat` to filter just the list of keys for the current category (would be surprising to copy things that you can't select/deselect by checkbox)
- check what happens on category state change - should reset to the new set of visible fields

app/src/frontend/building/data-container.tsx
```diff
@@ -80,11 +81,18 @@ const withCopyEdit: (wc: React.ComponentType<CategoryViewProps>) => DataContaine
         static getDerivedStateFromProps(props, state): DataContainerState {
             const newBuildingId = props.building == undefined ? undefined : props.building.building_id;
             const newBuildingRevisionId = props.building == undefined ? undefined : props.building.revision_id;
+            const categoryKeys = {};
+
+            for (let key in dataFields) {
+                if (dataFields[key].category == props.cat && props.building[key] != null){
+                    categoryKeys[key] = true;
+                }
+            }
```
edwardchalstrey1, line 90: Hi @tomalrussell I've implemented your idea for filtering on category and only selecting fields that are filled within `getDerivedStateFromProps`

The part that doesn't work currently is `props.building[key]`. The props.building allows me to use a key that comes from the Building interface in `frontend/models/building.ts` e.g. `revision_id`, but not a key from `frontend/config/data-fields-config.ts` e.g. `date_year`, which I would have expected to be the case. When I `console.log(props.building)` it looks like a dictionary structure where all the keys are at the same level. Any thoughts?

edwardchalstrey1: Ok think I've got a workaround for this actually

edwardchalstrey1:
Categories with copying issues remaining:
- [ ] Current Use
- [ ] Planning
- [ ] Community

PR #843, opened 2022-06-24
Rename "Streetscape" to "Street Context"

edwardchalstrey1:
Hi @tomalrussell @matkoniecz could either of you help with this? It seems like it should be a simple thing, but I'm not sure how to edit the category names at the front end, is there a somewhere in the codebase where the displayed names can be edited?

e.g. it should look like this after the change:
Screenshot 2022-06-24 at 10 14 31


matkoniecz:
I am looking at code and do not see anything obvious. Is it possible that it was simply cached and you need to clear browser cache?

edwardchalstrey1:
Thanks @matkoniecz you were absolutely right!

PR #878, opened 2022-07-22
Fix counting verifications

### LINKED ISSUE #855: Fix verify counter
Problem:
The verification counter does not increment when multiple users verify the same field, it gets stuck on 1 or 0.

tomalrussell:
Closes #855 - issue seems to be in the loop which counts up the verifications for each attribute.


edwardchalstrey1 (approved):
Thanks Tom, works for me 👍

PR #885, opened 2022-08-09
Add ways of working diagrams

edwardchalstrey1:


edwardchalstrey1:
This was added to the manual so don't think we need it here too

PR #897, opened 2022-08-30
Feature/display of planning data

matkoniecz:
@edwardchalstrey1 If you would have time to review it, even in part: I would be grateful!

Issue #884, opened 2022-08-01
Planning visualisation tool questions


matkoniecz:
# Questions

- [x] Section Title (under Planning category): _Planning Application Status_
- [x] Add Disclaimer - to be agreed 

**In  Green box** **(automated)**
- [x] Add title for box  _'Local Authority planning information'_ and add items below 
- [x] Add _'Data Source'_: The Planning London DataHub Greater London Authority plus hyperlink 
- [x] Add _'Planning Application ID'_
- [x] Add '_Most recent update by data provider_'
- [x] Add _'Planning Portal link'_ from url_planning_app field - if not available should say 'link not provided'
- [x] Add '_Current planning application status for this site_':
- [x] Add either _'Submitted', 'Approved', 'Rejected', 'Withdrawn', 'Appealed'_
- [x] Add _'Date'_
- [x] Add _'Brief Description of proposed work'_ (max 400 characters) plus _'Show more...'_
- [ ] Add Development constraints (see issues below)
- [ ] Add Conservation area layer on map (Ian Hall dataset) 
    - option to switch on/off

**Out of Green box (crowdsourced)**
- [ ] Add _'Do you think a planning application _is going to_ be submitted for this site that is not currently visualised?'_ 
- [ ] Add _'Edit date'_ - link this to edit history/automated
(crowdsourced)
- [ ] Add _'Are you aware of a planning application that has been _recently_ submitted for this site and is not listed below?'_ 
- [ ] Add _'Edit date'_ - link this to edit history/automated
- [ ] Add _'Has the work on this site been completed?'_ (crowdsourced)
- [ ] Add  _'Estimated date of completion'_ (crowdsourced- link this to age section to autopopulate 'latest possible construction date)

**Issues/Comments**
- Protections/permissions needed e.g. Listed buildings consent / conservation area consent (should be able to see on map if required / empty box), etc. Also areas of archaeological priority, SSSI, local listed buildings (Ian Hall developing dataset) - to be discussed

- Question on demolition 'Do you expect this building to be demolished?'
Preferable to show demolition question both in "Community" and "Planning" section
- Multiple planning applications can apply to a single building, including multiple active applications and visualise latest
- Colours for specific planning stages can be changed

fallipalaiologou:
@polly64 we currently call the section 'Planning Status' see [here](https://colouringlondon.org/view/planning/2478530). Should we leave as is or rename to 'Planning Application updates' ?

polly64:
[...]

PR #898, opened 2022-08-30
Feature/display of planning data

matkoniecz:
replaces #897 due to change to permission levels


polly64:
Hi @edwardchalstrey1  @fallipalaiologou 
Hi @matkoniecz
please can you also

- take old questions down  
- change link in top box  to the https://github.com/colouring-cities/manual/wiki/G1.-Data-capture-methods
- Change disclaimer text based on my earlier comments thanks
-remove the title 'Planning application status' but keep the same title where it exists  with the arrow next to just put green box within concertina 
- Give a planned date for activation of the out of green box crowdsourcing questions- Can you send these to ed today? i.e. at  https://github.com/colouring-cities/colouring-london/issues/884. 
he is leaving and there will be a delay before Mike starts and he will have a lot of other priorities 

- Also a lot of the entries are yellow which means appeals but these are normally rare- something wrong here could you check thanks

matkoniecz:
@tomalrussell Review is also welcome

etl/planning_data/obtain_livestream_data_and_load_into_database.py
```diff
@@ -0,0 +1,310 @@
...
+        '_source': {
+            'excludes': [],
+        },
+        'query': {
+            'bool': {
+                'must': [
+                    {
+                        'range': {
+                            'valid_date': {
+                                'gte': '01/01/1021',
```
tomalrussell, line 139: 1021 is a long time ago - was that meant to be 2021?

etl/planning_data/obtain_livestream_data_and_load_into_database.py
```diff
@@ -0,0 +1,310 @@
...
+import datetime
+import os
+import requests
+import psycopg2
+import address_data
+
+def main():
+    connection = get_connection()
+    cursor = connection.cursor()
+    cursor.execute("TRUNCATE planning_data")
```
tomalrussell, line 11: Is this meant to be re-run regularly? Could we do something more conservative than `TRUNCATE` and reload all? It seems quite risky to drop everything and hope the bulk reload works okay every time.

etl/planning_data/obtain_livestream_data_and_load_into_database.py
matkoniecz: In this case it was intended to cover all data (as one point I tested it and it needed some minimum date).

There is scattering of some old data like something from 1968, so in theory they may add older ones. Not sure when the first equivalent of planning permission appeared in London, so I set date that would catch everything.

matkoniecz: Though I will add comment explaining situation and retry is this weird filter avoidable.

mdsimpson42:
I've checked it out and it looks fine to me.

matkoniecz:
@tomalrussell Is it safe (and a good idea) to store in `geometries` table geometries not representing buildings?

It seems to be necessary to display borough labels requested by @polly64 - because different boroughs have drastic differences in data quality they provide.

And to get decent labels I render them using mapnik, with one more tile layer, as that seems beyond what GeoJSON layer can do.

It would be also possible to put them in a completely separate table, without putting this data into geometries table.

main commit is in https://github.com/colouring-cities/colouring-london/pull/898/commits/c4933fdc11118e78d19060584280304485c613c0 (with some later tweaking of label styling)

matkoniecz:
Also, I keep wondering what is the best place to store geometry migration data - I am not really happy about https://github.com/colouring-cities/colouring-london/pull/898/commits/c6831b9f794e698ca95acec15f1e90760ac2aaa9 putting them in migration file, but it would be nice to have them somewhere (it was already a bit tricky to reconstruct them for me, now just few days after initial coding and being the author as I deploy them to staging)

matkoniecz:
@mdsimpson42 This one should be overall in mergeable state. There are some minor upcoming tweaks on TODO list, but nothing broken/fundamental.

(though I am sure that review would find things that could be done better)

Issue #933, opened 2022-09-22
Include autogenerated specific entry listings in Planning


matkoniecz:
@polly64 

Can you try to track down links to specific entry listings for 

 - [x] "Is the building on the [National Heritage List for England](https://historicengland.org.uk/advice/hpg/heritage-assets/nhle/)?" - https://historicengland.org.uk/listing/the-list/list-entry/1080446?section=official-list-entry
 - [ ] "Is it on the [Heritage at Risk](https://historicengland.org.uk/advice/heritage-at-risk/search-register/) register?"
 - [ ] "Is it a [Locally Listed Heritage Asset](https://historicengland.org.uk/advice/hpg/has/locallylistedhas/)?"
 - [ ] "Is it within a [World Heritage Site](https://historicengland.org.uk/advice/hpg/has/whs/)?"
 - [ ] "Is it recorded on the [Historic Environment Record](https://historicengland.org.uk/advice/technical-advice/information-management/hers/)?"
 - [ ] "Is it in an [Archaeological Priority Area](https://historicengland.org.uk/services-skills/our-planning-services/greater-london-archaeology-advisory-service/greater-london-archaeological-priority-areas/)?"
 - [ ] "Does it have an [Historic Area Assessment](https://historicengland.org.uk/images-books/publications/understanding-place-historic-area-assessments/)?"

?

matkoniecz:
What is the id format: pure numbers or is say 106a also a potential id?

matkoniecz:
@polly64 

https://cl-staging.uksouth.cloudapp.azure.com/edit/planning/2956004 - can you simply give example of links that user would put as answer, in currently deployed staging version. That would answer my need for "links to specific entry listings"

matkoniecz:
@polly64 

Can you write text for "How to find iD" minitutorials?

matkoniecz:
Ideally this boxes here would narrow, with link on the right side. Not promising this version, will try to implement

![Untitled](https://user-images.githubusercontent.com/899988/191785103-a13b56ae-a058-4903-9360-2bcbecd11953.png)

matkoniecz:
@polly64 

It looks like this:

![screen27](https://user-images.githubusercontent.com/899988/191978647-4ec11ff7-64b9-4c85-b34f-d462bde21556.png)

polly64:
[...]

PR #941, opened 2022-09-24
More uniform form style

### LINKED ISSUE #937: Some forms do not have rounded corners - is it deliberate?
@polly64 

Or should it be changed so all forms have round corners?

screen28

matkoniecz:
fixes #937

maybe I missed some unexpected effects but it seems to be not needed and just increases undesirable clutter

see https://github.com/colouring-cities/colouring-london/issues/933#issuecomment-1256269206


matkoniecz:
@mdsimpson42 This one is more simple and also waits for review

PR #943, opened 2022-09-25
fix off by one error in username legth limit

matkoniecz:
note

> ALTER TABLE users ADD CONSTRAINT users_username_len CHECK (length(username) < 30);

PR #956, opened 2022-10-12
Tweak login menu messages

matkoniecz:


matkoniecz:
@tomalrussell @mdsimpson42 Can you review this one? It is relatively simple and obvious, @polly64 was asking also for extra check of registration, but I think that I solved this in #943

PR #975, opened 2022-11-08
Stop showing child count in group headers.

matkoniecz:
as requested @polly64


matkoniecz:
@mdsimpson42 Can you review this one? It is a relatively simple change

Issue #1009, opened 2022-12-23
Planning section deployment


matkoniecz:
- [x] process reviews if any
- [x] merge PR https://github.com/colouring-cities/colouring-london/pull/898 (can be done by @mdsimpson42 or @tomalrussell )
- [x] deploy on the main site additional data layers
- [ ] monitor automatic streaming after merge
- [x] merge and deploy on the main site planning data

matkoniecz:
was part of #1002

polly64:
@matkoniecz not seeing borough boundaries?

matkoniecz:
I will do this once I got back from trip (on 3rd January)

polly64:
many thanks

matkoniecz:
@polly64 Do you want to keep size from the second image on the third one?

Or do you wish to adjust text size at some other zoom level?

![screen01](https://user-images.githubusercontent.com/899988/210360731-d111874c-d7a3-4424-a2d8-fb06e38203df.png)
![screen02](https://user-images.githubusercontent.com/899988/210360728-7583e011-d351-420d-ba76-13e084126454.png)
![screen03](https://user-images.githubusercontent.com/899988/210360722-c5091579-ad06-4dae-ba2a-f35052386d21.png)

matkoniecz:
@polly64 

>  on Location section change 'Building name information (link)' to 'Building Name (Information link)

Do you mean that field?

![screen04](https://user-images.githubusercontent.com/899988/210373522-8c3abe7b-101e-4ba2-99cd-2ce3811bfb19.png)

matkoniecz:
@polly64 

>  on 'Search feature' allow you to search area without double clicking the postcode image attached

So you would want to move map to the first postcode on the list, if matching, on clicking "search" button?

matkoniecz:
@polly64 

> do you understand why there s a dropdown?

probably to help in cases where matches are not exact? Though it makes far less sense with postcode-only search, it is far more useful when you can search by name or part of the name - where typing only part of name/address is far more likely to be useful.

I guess that someone used this because this was the standard search menu or maybe it was planned/expected that general search by name will be added in the future.

matkoniecz:
>  Yes for location issue it is just to change the text where it currently says' building information link.

It seems to change meaning of the field - right now it seems to allow giving any website where more info will be found

matkoniecz:
so should I rename or hide that field or remove it entirely?

polly64:
[...]

PR #1014, opened 2023-01-03
rename field in Location section

matkoniecz:
requested in https://github.com/colouring-cities/colouring-london/issues/1009#issuecomment-1369811001 https://github.com/colouring-cities/colouring-london/issues/1009#issuecomment-1369910495


matkoniecz:
screen

PR #1019, opened 2023-01-04
set message about planned layer to yellow

matkoniecz:
for conistency with other similar boxes

### DIFF
--- a/app/src/frontend/building/data-containers/streetscape.tsx
+++ b/app/src/frontend/building/data-containers/streetscape.tsx
@@ -13,7 +13,7 @@ import { CategoryViewProps } from './category-view-props';
 */
 const StreetscapeView: React.FunctionComponent<CategoryViewProps> = (props) => (
     <Fragment>
-        <InfoBox msg="This is what we're planning to collect on the building's context" />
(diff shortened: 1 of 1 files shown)


matkoniecz:
screen

polly64:
@matkoniecz what are you using this one for?

polly64:
@matkoniecz would be good to be consistent based on planning colours and  then change other ones. So pink button brings layer up, blue button provides background . Pale grey with italics is disclaimer . We used yellow for 'not yet activated' but we could just use greyed out

matkoniecz:
> what are you using this one for?

I do not understand the question here

mdsimpson42:
@polly64 @matkoniecz Do we have approval for the colours etc. here? Or are you still discussing it?

matkoniecz:
this one was discussed at the meeting and yes, it should be yellow

polly64:
@mdsimpson42 @matkoniecz . Mateussz shouldn't it be blue because Explanatory info?

blue button provides background . Pale grey with italics is disclaimer . We used yellow for 'not yet activated' but we could just use greyed out

polly64:
ah ok

PR #1034, opened 2023-01-10
adjust flex to work better with long titles

matkoniecz:
screen07
screen06

tested with current troublesome ones, and with even longer ones (I assumed that in case of extra-long titles it would be better to wrap title than action labels)


matkoniecz:
@mdsimpson42 Can you review and maybe merge this (and other PRs)?

PR #1035, opened 2023-01-10
proper check on login

### LINKED ISSUE #1024: White screen (uncaught error) on login
**Describe the problem**
Logging in to the site leads to a white screen. Refreshing the page leaves the user successfully logged in.

**To Reproduce**
Steps to reproduce:
1. Go to colouring.london
2. Click on the Menu > Log In
3. Enter username and password correctly
4. Click 'log in'
5. See complete white screen
6. Back button does nothing
7. Refresh button is a successful workaround

**Expected behavior**
Log in successfully, get redirected to account page.

Errors should also be handled with a react error boundary, so we can at least show a message and suggest refreshing the page or linking back to the homepage.

**Logs**

Error in console:
```
TypeError: right-hand side of 'in' should be an object, got null
    render route.tsx:48
    n.prototype.render/$< react-router.js:445
    React 6
    unstable_runWithPriority scheduler.production.min.js:18
    React 5
    d auth-context.tsx:35
react-dom.production.min.js:216:199
```

matkoniecz:
fix applied by @tomalrussell works in testing on staging

fixes #1024

### DIFF
--- a/app/src/frontend/route.tsx
+++ b/app/src/frontend/route.tsx
@@ -39,7 +39,7 @@ export const AuthRoute: React.FC<RouteProps> = ({ component: Component, children
             if(isAuthenticated) {
                 let state = props.location.state as any;
                 let from = '/my-account.html';
-                if (typeof state == 'object' && 'from' in state){
+                if (typeof state == 'object' && state !== null && 'from' in state) {
                     from = state.from;
                 }
 


tomalrussell (approved):

tomalrussell:
Thanks for picking this up and testing.

PR #1061, opened 2023-01-19
Various tweaks discovered during testing of new layer

matkoniecz:


matkoniecz:
The biggest change/regression/hack is https://github.com/colouring-cities/colouring-london/pull/1061/commits/d73e7271dc66ba185fb21050332d4745f7ec51c1

For reason that are utterly unclear it noticeably reduces cases where loading of tiles is stuck. It is unclear why this effect is present - it should request more tiles with retina support, retina tiles have separate cache but still drastic difference is surprising and needs to be investigated.

Issue #1062, opened 2023-01-20
is allLayersCacheSwitch and dataLayersCacheSwitch even needed?


matkoniecz:
Do we need support for CACHE_TILES so that caching can be disabled by changing environment variable? And similar CACHE_DATA_TILES.

https://github.com/colouring-cities/colouring-london/blob/7199c7f6a3a7c0dbac4cd021a1d595cd0a99a571/app/src/tiles/rendererDefinition.ts#L28

That seems just bring extra complexity and I am not sure is it something that we need to have built-in special reconfiguration and makes a bit harder to reason what is going on.

Opening an issue as I do not want to just remove things I am not understanding.

tomalrussell:
I think the use case for these variables is to help with the development of new tile styles: disabling the cache would make it easier to check and test changes to design rules. 

On balance, I'd recommend leaving it there (maybe add a little doc comment), as long as it's not causing any other issues.

matkoniecz:
then I need to check is it also affecting retrieval from cache. Is it needed to delete caches and flip this variable - or is it enough to just flip this variable - to test layers without caching.

(personally, just emptying cache folder works fine for me)

polly64:
@matkoniecz can we close

PR #1063, opened 2023-01-20
more aggressive caching

matkoniecz:
in tests it seems confirmed that caching noticeably improves performance (Polly noticed that age layer is noticeably faster without being aware that it is cached more aggressively, in testing in general all layers benefit from caching even at zoom level 18 for relatively minor storage cost and differences in the first use performance that are basically not possible to measure (or at least I failed to measure them)

caching will be especially important as more people will use the site and will help reduce server costs

see also #1062 proposing further simplification of this code


matkoniecz:
@mdsimpson42

PR #1082, opened 2023-02-09
make data entry group work with long titles

matkoniecz:
may have some unexpected side-effects but testing has not revealed any


matkoniecz:
Before this change long header text was overlapping with other content, after this change it is not happening anymore.

Such long text is can and is and should be avoided in English, but would potentially cause problems in some rare cases or with future potential translations.

PR #1128, opened 2023-04-06
Feature/survival status

mdsimpson42:


mdsimpson42:
Polly wants to test this feature, even though it's still a work in progress. Bug fixes will be applied later.

matkoniecz:
Note that in such case it is possible to deploy branch on staging without merging it already.

mdsimpson42:
True, but it was developed in Core and needs to be in London for deployment... I don't want to mess around too much changing origins at this point either (especially at 4pm on the Thursday before Easter!).
 
I agree, I need to find a better way to do it! Perhaps something to be discussed later.

PR #1217, opened 2023-07-27
Revert "Add News and Discussions Links"

mdsimpson42:
Reverts colouring-cities/colouring-core#1216


mdsimpson42:
Supposed to be on London, not Core

Issue #1284, opened 2023-12-13
Add map layer url APIs to and tile render bounding box to config


Ruben1729:
# Problem 
While trying to port colouring-core to the city of Montreal, our team encountered a few problems. These are not documented (at least not in the [dev environment docs](https://github.com/colouring-cities/colouring-core/blob/master/docs/setup-dev-environment.md) or any other docs that we could find).

The map layer does not render outside the UK. Additionally, even after changing the map layers, the buildings do not render outside London.

# Solution
To make it work we had to change 4 files:

In `colouring-core/app/src/frontend/map/layers/city-base-map-layer.tsx` we replaced this
```tsx
const baseUrl = `https://api.os.uk/maps/raster/v1/zxy/${layer}/{z}/{x}/{y}.png?key=${apiKey}`;
```
with
```tsx
const baseUrl = `https://a.tile.openstreetmap.org/{z}/{x}/{y}.png`;
```

In `colouring-core/app/src/frontend/map/layers/historic-map-layer.tsx` we replaced this
```tsx
<TileLayer url="https://mapseries-tilesets.s3.amazonaws.com/london_1890s/{z}/{x}/{y}.png" />
```
with
```tsx
<TileLayer url="https://a.tile.openstreetmap.org/{z}/{x}/{y}.png" />
```

In `colouring-core/app/src/frontend/map/layers/historic-data-layer.tsx` we replaced this
```tsx
<TileLayer url="https://mapseries-tilesets.s3.amazonaws.com/london_1890s/{z}/{x}/{y}.png" />
```
with
```tsx
<TileLayer  url="https://a.tile.openstreetmap.org/{z}/{x}/{y}.png"/>
```

In `colouring-core/app/src/tiles/rendererDefinition.ts` we replaced this
```ts
const EXTENT_BBOX: BoundingBox = [-61149.622628, 6667754.851372, 37183, 6744803.375884];
```
with
```ts
// Montreal bounding box
const EXTENT_BBOX: BoundingBox = [-8135086.254447, 5675540.368758, -8267367.825970, 5751427.756290];
```

It would be useful to update the [configuring-colouring-cities.md](https://github.com/colouring-cities/colouring-core/blob/master/docs/configuring-colouring-cities.md) to describe how to change the bounding box and map layer URLs.

# Alternatives
Update the config file to account for these links as well as the bounding box to allow developers to easily change the map layer API and bounding box to fit their city.

matkoniecz:
[...]

PR #1286, opened 2023-12-18
Add configuration option for tile bbox

matkoniecz:
partially addresses #1284

tested with `"bbox": [-13821, 6707971, -7886, 6711363]`

screen06

### DIFF
--- a/app/src/cc-config.json
+++ b/app/src/cc-config.json
@@ -10,5 +10,7 @@
     "initialZoomLevel": 16,
 
     "postcode": "Postcode",
-    "energy_rating": "Building Research Establishment Environmental Assessment Method (BREEAM) rating [<a href='https://bregroup.com/products/breeam/how-breeam-works'>More info</a>]"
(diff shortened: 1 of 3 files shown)


matkoniecz:
I considered also more verbose config format, with values like `tile_bbox_epsg_3857_south_limit`, `tile_bbox_epsg_3857_north_limit`, `tile_bbox_epsg_3857_east_limit`, `tile_bbox_epsg_3857_west_limit`

matkoniecz:
@Ruben1729 review is welcome!

KoaCWells:
@matkoniecz May I ask what the intention of the bounding box is? From my impression so far, each Colouring City project should only have building polygons that belong to their corresponding city so I am having a bit of trouble understanding why it is used. Is there some sort of benefit in terms of performance or necessity? We were able to remove the bounding box entirely from Colouring Montreal without any issues so far. Thanks in advance for your feedback!

matkoniecz:
@KoaCWells Without this bounding box panning across world will create empty tiles that will be then cached.

Maybe extra complexity added by this option and small space saved is making it not worth it? It may be useful to check how much space is saved in the best case (someone dragging map across entire world, at every zoom level).

PR #1297, opened 2024-01-09
Allow easier change of map tiles and attribution, use tiles with global coverage

matkoniecz:
Move basemap layer configuration to configuration file.

fixes https://github.com/colouring-cities/colouring-core/issues/1284

Use OpenStreetMap tiles by default.

More friendly to new partners outside UK, less configuration will be typically required.

fixes https://github.com/colouring-cities/colouring-core/issues/1281

Tested in UK, but we will almost certainly stay with OS basemap there (needs to be taken into account when merging upstream - https://github.com/colouring-cities/colouring-core/commit/7ef7fedc3a6fc74262c5c3a580d6d0fb139fd0b9 should be partially reverted then).
screen03

Test with Acropolis of Athens (no loaded buildings)
screen04

Day palette is a bit noisy, maybe colour transformation to desaturate/make colours gray would be a good idea? (like dark mode is achieved, map tiles are the same). But that would be a separate PR.

screen05

### DIFF
--- a/app/src/cc-config.json
+++ b/app/src/cc-config.json
@@ -14,6 +14,6 @@
 
     "bbox": [-61149.622628, 6667754.851372, 37183, 6744803.375884],
 
-    "basemapTileUrl": "https://api.os.uk/maps/raster/v1/zxy/Light_3857/{z}/{x}/{y}.png?key=UVWEspgInusDKKYANE5bmyddoEmCSD4r",
-    "baseAttribution": "Building attribute data is © Colouring Cities contributors. Maps contain OS data © Crown copyright: OS Maps baselayers and building outlines. <a href=/ordnance-survey-licence.html>OS licence</a>"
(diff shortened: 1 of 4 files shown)


matkoniecz:
@polly64 To confirm that it is fine to change default basemap (I assume yes, based on discussion from international meeting)

matkoniecz:
screen06

on, and default attribution was modified to attribute new default basemap and with placeholder for building source

matkoniecz:
Overly cautious conflict of interest note: I am deeply involved in OSM community. Though note that this was also independently suggested/done in #1284 and there are no obvious alternatives.

polly64:
@matkoniecz shall I move back to core?

polly64:
@matkoniecz can't see base map images can u resend but yes if agreed by others

mdsimpson42 (approved):
Looks fine to me. Happy to merge and then we can deal with any unforeseen issue.

PR #1319, opened 2024-01-16
Update node version in GitHub Action

mdsimpson42:
Attempt to fix failed run in https://github.com/colouring-cities/colouring-core/pull/1275

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x]
+        node-version: [15.x]
 
     steps:
     - uses: actions/checkout@v2


matkoniecz:
@mdsimpson42 

Note that on production server we have `node --version` giving `v12.22.12` so if #1275 breaks node v12 then we need to upgrade it on servers (and test is site in general still working)

mdsimpson42:
Ah, OK. Yes, it's the same on Staging, too. I'll test it on there first, but it passed the tests without any issues and works locally 🤞

mdsimpson42:
I installed NVM on Staging, then installed Node v15.0.0 and built the app from the latest code and everything seems to work. I'll do the same on Production next time we do a new deployment.

matkoniecz:
Are you planning it soon? Maybe redeploy with this change and new node to confirm that everything is fine?

(if something would go wrong in future it could be more complicated to track down what went wrong)

PR #1320, opened 2024-01-16
Remove unused geometry file.

matkoniecz:
also, it was London-specific anyway.


matkoniecz:
`rg "boundary-detailed.geojson"` finds

```
app/src/frontend/map/layers/city-boundary-layer.tsx
11:        apiGet('/geometries/boundary-detailed.geojson')
```

`rg "boundary.geojson"` finds nothing.

Tested in VM with file removed, everything works fine.

@polly64 Do we need this file for anything in core repository? Should it be also removed from Colouring Britain?

Geometries are visualised currently at https://github.com/colouring-cities/colouring-core/blob/master/app/public/geometries/boundary-detailed.geojson (for used file) and https://github.com/colouring-cities/colouring-core/blob/master/app/public/geometries/boundary.geojson (for one being proposed to be removed here)


(spotted while adding display of more geometries for Colouring Britain)

PR #1328, opened 2024-02-06
Do not assume that all platforms are about London.

matkoniecz:
@polly64 

Should we remove 

> We’re building it at The Bartlett Centre for Advanced Spatial Analysis, University College London.

without replacement or change it?


polly64:
@matkoniecz yes please

matkoniecz:
so this mention should be remove without replacement, right?

matkoniecz:
see also https://github.com/colouring-cities/colouring-britain/issues/280

PR #1350, opened 2024-03-14
Revert "Add sample historical footprints"

mdsimpson42:
Reverts colouring-cities/colouring-core#1349


matkoniecz:
This adds empty commit.

I would consider undoing merge by restoring previous branch state and force pushing this changes.

PR #1360, opened 2024-03-28
Remove Newcastle University logo

matkoniecz:
see https://github.com/colouring-cities/colouring-britain/issues/318


mdsimpson42:
If we're removing Newcastle from Core, should we remove some of the other logos too? It might make sense to remove all of them except for the CCRP and Turing logos? (They're all London/Britain-specific apart from those two)

Are we keeping them on Colouring Britain?

Thoughts @polly64?

matkoniecz:
https://github.com/colouring-cities/colouring-britain/issues/318#issue-2204083337 has

>  Hide Newcastle university logo - university logos are only included where free resources and/or research time have been contributed by a university

PR #1368, opened 2024-04-01
Rename section

matkoniecz:
from "Open property coordinates" to "Base building footprints/polygons used for data collation, verification and visualisation"

see https://github.com/colouring-cities/colouring-britain/issues/275#issuecomment-2026892424

screen

@polly64


mdsimpson42:
My only comment is that it's quite a long heading. I believe it's the only subcategory longer than a single line. There may be a way to shorten the actually heading (i.e. to just "Base building footprints/polygons") and add some descriptive text (i.e. the "used for data collation, verification and visualisation" bit) underneath the heading? Just a thought @polly64. Will merge for now anyway.

PR #1376, opened 2024-05-24
update npm version for doc set up dev env

traveller195:
to avoid error of wrong npm version.. using npm@8.1.2 instead of latest


matkoniecz:
thanks so much for testing and PR!

what kind of errors appears? Maybe we can update things so will not really on ancient npm version?

mdsimpson42:
Closing. The documentation in question has been moved to the [Technical Manual/Wiki](https://github.com/colouring-cities/ccrp-technical-manual/wiki). That documentation appears to have been updated, albeit with a slightly different version of npm

PR #1398, opened 2024-09-25
Lift length restrictions on planning data strings.

matkoniecz:
See https://wiki.postgresql.org/wiki/Don%27t_Do_This#Don.27t_use_varchar.28n.29_by_default

Import started to fail as at least one link is exceeding 260 characters, limit that seemed reasonable. Lets lift this limits in general here.


matkoniecz:
note: downgrade will fail with `ERROR:  value too long for type character varying(260)` or similar if long value is present

I assume that it is fine

mdsimpson42 (approved):

PR #1416, opened 2024-10-10
Interface Updates

mdsimpson42:
- Swap Infrastructure Categories around on the menu
- Tweaks to Common Source Types list
- URL Fixes and remove Britain-specific content


mdsimpson42:
These are minor changes, requested urgently. I've tested them and will test them again before merging into Colouring Britain and deploying.

PR #1420, opened 2024-10-16
Do not claim that location_longitude location_latitude is centroid

matkoniecz:
For L shaped buildings it is better to select location that is within building, not a centroid.


matkoniecz:
@polly64 This changed text will be shown on the location page

PR #1434, opened 2024-11-20
Now only legend is hidden in basic view mode.

matkoniecz:
as requested by @polly64


matkoniecz:
test is failing but it is a problem with test

PR #1437, opened 2024-11-20
Revert "Merge pull request #1434 from matkoniecz/feature/show_more_interface_elements_by_default

matkoniecz:
This reverts commit 7ab2edeba948346c3e0257f9d4bd83ca6189d2fe, reversing changes made to bb7ba04beff995dd708cd305e54ef0255220f418.


matkoniecz:
It looks like always showing ThemeSwitcher, SearchBox, DataLayerSwitcher and other switchers is causing problems in tests.

Some of these elements are using `fetch`. Tests are failing now as these elements are now shown by default.

Note: I will try to fix it rather than merge this, as reverting merge commit results in really weird things.

matkoniecz:
TODO: test which elements causes failure

TODO: try to mock fetch in the test `src/frontend/app.test.tsx`

PR #1439, opened 2024-11-21
More prominent "start colouring" button

matkoniecz:


matkoniecz:
as requested in https://github.com/colouring-cities/colouring-britain/issues/378#issue-2652336519

PR #1445, opened 2024-11-25
now 'Current Building on Previously Unbuilt-on Site'

matkoniecz:
@polly64 Can you confirm "Current Building on Previously Unbuilt-on Site" as final text? This is a bit fiddly to migrate.

Also, needs a bot (requires manual run by other instances of the platform) or database migration script which will make history confusing.

### DIFF
--- a/app/map_styles/polygon.xml
+++ b/app/map_styles/polygon.xml
@@ -937,7 +937,7 @@
             <PolygonSymbolizer fill="#ff2121" />
         </Rule>
         <Rule>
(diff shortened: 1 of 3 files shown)


polly64:
@matkoniecz yes but maybe we can lose the capital P on previously

polly64:
@matkoniecz or let me know your preference? is this not just a word change?

matkoniecz:
> is this not just a word change?

the problem is with already existing data that needs to be migrated

PR #1446, opened 2024-11-25
Remove tooltips unrelated to these fields.

matkoniecz:
@polly64 

Should 

> "Cladding date (best estimate)"

> "Date of significant extensions (best estimate)"

> "Date of last significant retrofit (best estimate)"

should get replacement tooltips?

This change removes "Width of the street in metres." that is their current unwanted tooltip.


polly64:
@matkoniecz do you want to come on a call now so we can go through the list?

matkoniecz:
@polly64 These tooltips are now properly gone.

If you see any other that should be silenced, please create an issue.

PR #1448, opened 2024-11-25
Tweak landing page text.

matkoniecz:
@polly64 - for review

I will try to rearrange logos now. It will be a bit irritating but should be technically simple.

screen07

### DIFF
--- a/app/src/frontend/pages/welcome.css
+++ b/app/src/frontend/pages/welcome.css
@@ -29,7 +29,13 @@
     height: 4em;
 }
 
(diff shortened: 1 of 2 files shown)


matkoniecz:
It is here Colouring Cities, not Colouring Britain as I edited base text for all platforms.

Should logos be changed in core or for Britain only? I guess that core, as it also helps develop core platform.

polly64:
only Colouring Britain for moment please but leave Turing and EPSRC on the others thanks,

PR #1455, opened 2024-12-04
Simplify landuse autogeneration message.

matkoniecz:
before:

screen12

after:

screen13

implements part of https://github.com/colouring-cities/colouring-britain/issues/391


matkoniecz:
replaced by #1459

PR #1456, opened 2024-12-04
Improve survival_source field based on feedback from historians.

matkoniecz:
implements part of https://github.com/colouring-cities/colouring-britain/issues/391


matkoniecz:
before:

screen14

after:

screen15

PR #1459, opened 2024-12-04
Feature/reorganize landuse

matkoniecz:


matkoniecz:
@polly64 - it is result of review yesterday. It does not include all changes, especially ones requiring more involved coding, but easy to implement ones are done. Do you see any mistakes?

screen07

PR #1479, opened 2024-12-12
Remove some unhelpful tooltips.

matkoniecz:


matkoniecz:
example - current state on the right

this has appeared on one of meetings - pinging @polly64 to confirm that I understood correctly, you can also look at https://github.com/colouring-cities/colouring-core/pull/1479/files to see (red on the left) what kind of tooltips got removed

@polly64 - I can spend some time on that and review all the tooltips (or some parts of them) if it is worth spending time on it. Let me know if you decided either way.

screen

matkoniecz:
from meeting, to be applied to source type tooltips in general:

"method used to collect data"

PR #1491, opened 2025-03-27
Testing to see if this fixes an error in GitHub actions

mdsimpson42:


mdsimpson42:
@matkoniecz I updated the version of `cache` to avoid the error that was happening in this branch https://github.com/colouring-cities/colouring-core/pull/1490. It seems to have gotten rid of the error message, but I'm not sure if this affects anything else...

mdsimpson42:
On the basis that these are just checks that get run when the code is checked in, and don't have anything to do with deployment, I am going to merge these changes.

PR #1495, opened 2025-05-02
Feature/easy redesign

matkoniecz:


matkoniecz:
@polly64 This is now on staging :)

Together with already merged tooltip cleanup.

Note that it is only part of TODO list from recent meetings, but these changes are among easier ones and self-contained.

PR #1500, opened 2025-05-14
move also community_local_significance_total map, like question was moved

matkoniecz:


matkoniecz:
https://github.com/colouring-cities/colouring-core/pull/1501

PR #1513, opened 2025-09-05
Exagerrate building size for EPC layer to make data more visible on lower zoom levels.

matkoniecz:
z11 before/after
z11 before
z11 now

z13 before/after
z13 before
z13 now

### DIFF
--- a/app/map_styles/polygon.xml
+++ b/app/map_styles/polygon.xml
@@ -2036,37 +2036,39 @@
         <Rule>
             <Filter>[sust_aggregate_estimate_epc] = A</Filter>
             <PolygonSymbolizer fill="#007f3d" />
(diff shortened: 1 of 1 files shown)


matkoniecz:
change is fairly simple: draw not only building area but also its border

it works as building may be sized as 1 pixel (or even 1/10 of a pixel!) so fill will draw pale, barely noticeable mark

while 3px-sized border will produce such border also when building is tiny tiny

there are also negative effects (exaggeration of tiny buildings, one building may block another, distortion of building shapes etc) but they should be outweighed by increased readability and being able to actually see data, especially on screen worse than high-end ones in carefully lit rooms :)

draft as @polly64 requested to apply it more widely, not only to EPC test

polly64:
@matkoniecz let's do for everything

